PR #19, opened 2025-04-05
fix: simplify NodePoolTemplate provider by using ClusterInfo struct

patrostkowski:
#### What type of PR is this?

fix

This PR removes the dependency on existing NodePools by switching to direct GCP API calls for retrieving cluster and node pool information. This change enables more flexible and resilient bootstrapping workflows by eliminating the requirement for a default (empty) NodePool to exist in advance.

Additionally, this resolves a bug where templated NodePools would not run if a matching node template already existed. Reconciliation now works as expected in these scenarios.

Key updates:

- Introduced a ClusterInfo struct to encapsulate project, region, cluster name, and zones.
- These values are now configured via environment variables for easier setup.
- Refactored functions to accept ClusterInfo instead of individual parameters, improving code clarity and maintainability.
- Improved logging and error messages to enhance visibility and make debugging easier.
- Added example CRDs and Terraform files to assist with testing and setup.

none

none

none

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -15,6 +15,11 @@ KARPENTER_VERSION ?= $(shell git tag --sort=committerdate | tail -1 | cut -d"v" 
 KO_DOCKER_REPO ?= ko.local
 KOCACHE ?= ~/.ko
 
+# GCP Cluster Context
+PROJECT_ID ?= karpenter-provider-gcp
+CLUSTER_NAME ?= karpenter-provider-gcp
+CLUSTER_LOCATION ?= us-central1
+
 # Common Directories
 MOD_DIRS = $(shell find . -path "./website" -prune -o -name go.mod -type f -print | xargs dirname)
 KARPENTER_CORE_DIR = $(shell go list -m -f '{{ .Dir }}' sigs.k8s.io/karpenter)
@@ -35,6 +40,8 @@ run: ## Run Karpenter controller binary against your local cluster
 		KUBERNETES_MIN_VERSION="v1.26.0" \
 		DISABLE_LEADER_ELECTION=true \
 		CLUSTER_NAME=${CLUSTER_NAME} \
+		PROJECT_ID=${PROJECT_ID} \
+		CLUSTER_LOCATION=${CLUSTER_LOCATION} \
 		INTERRUPTION_QUEUE=${CLUSTER_NAME} \
 		FEATURE_GATES="SpotToSpotConsolidation=true" \
 		go run ./cmd/controller/main.go
--- a/Makefile
+++ b/Makefile
@@ -1,5 +1,3 @@
-CLUSTER_NAME ?= $(shell kubectl config view --minify -o jsonpath='{.clusters[].name}' | rev | cut -d"/" -f1 | rev | cut -d"." -f1)
-
 ## Inject the app version into operator.Version
 LDFLAGS ?= -ldflags=-X=sigs.k8s.io/karpenter/pkg/operator.Version=$(shell git describe --tags --always | cut -d"v" -f2)
 
@@ -18,7 +16,7 @@ KOCACHE ?= ~/.ko
 # GCP Cluster Context
 PROJECT_ID ?= karpenter-provider-gcp
 CLUSTER_NAME ?= karpenter-provider-gcp
-CLUSTER_LOCATION ?= us-central1
+REGION ?= us-central1
 
 # Common Directories
 MOD_DIRS = $(shell find . -path "./website" -prune -o -name go.mod -type f -print | xargs dirname)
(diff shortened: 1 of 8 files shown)


pkg/providers/nodepooltemplate/nodepooltemplate.go
```diff
@@ -70,109 +73,130 @@ func NewDefaultProvider(kubeClient client.Client, computeService *compute.Servic
...
+	logger := log.FromContext(ctx)
+
+	logger.Info("resolving zones", "ProjectID", p.ClusterInfo.ProjectID, "Region", p.ClusterInfo.Region)
+
+	var zones []string
+	err := p.computeService.Zones.List(p.ClusterInfo.ProjectID).Pages(ctx, func(page *compute.ZoneList) error {
 		for _, zone := range page.Items {
-			if strings.HasPrefix(zone.Name, region) {
-				ret = append(ret, zone.Name)
+			if strings.HasPrefix(zone.Name, p.ClusterInfo.Region+"-") {
```
jxs1211, line 167: Is moving the concat out of the loop would better?

patrostkowski: oh I think this part sneak to the code, indeed moving this out of the loop is better for less allocations

jxs1211:
LGTM

Makefile
```diff
@@ -15,6 +15,11 @@ KARPENTER_VERSION ?= $(shell git tag --sort=committerdate | tail -1 | cut -d"v"
 KO_DOCKER_REPO ?= ko.local
 KOCACHE ?= ~/.ko
 
+# GCP Cluster Context
+PROJECT_ID ?= karpenter-provider-gcp
+CLUSTER_NAME ?= karpenter-provider-gcp
```
jwcesign, line 20: Can we extract the cluster name from the existing nodes? I notice that we can retrieve it from the node labels using the key `cluster_name`.
```
labels:
    beta.kubernetes.io/arch: amd64
    beta.kubernetes.io/instance-type: e2-standard-4
    beta.kubernetes.io/os: linux
    cloud.google.com/gke-boot-disk: pd-standard
    cloud.google.com/gke-container-runtime: containerd
    cloud.google.com/gke-cpu-scaling-level: "4"
    cloud.google.com/gke-logging-variant: DEFAULT
    cloud.google.com/gke-max-pods-per-node: "110"
    cloud.google.com/gke-memory-gb-scaling-level: "16"
    cloud.google.com/gke-netd-ready: "true"
    cloud.google.com/gke-nodepool: cloudpilot-production-nodepool
    cloud.google.com/gke-os-distribution: cos
    cloud.google.com/gke-provisioning: standard
    cloud.google.com/gke-stack-type: IPV4
    cloud.google.com/machine-family: e2
    cloud.google.com/private-node: "false"
    cluster_name: cloudpilot-production
...
```

jwcesign (commented):
Thank you for your contribution! We look forward to your continued involvement in the GCP Provider community-building efforts!

Just a few minor points, otherwise LGTM!

Makefile
patrostkowski: hmm I just removed dependency of having non-empty nodepool by using GCP API directly. Reading those labels from any host creates dependency I think we want to aviod.

I think I will follow what is done here https://github.com/cloudpilot-ai/karpenter-provider-alibabacloud/blob/main/pkg/operator/options/options.go and add flags for cluster name, project ID and region instead.

jwcesign: With Alibaba Cloud, I noticed that we cannot resolve the cluster name from any resources.
However, with GCP, the situation is different. We can resolve the cluster name using VM labels
image

I prefer to request the least amount of input information from end users.

Note: Sorry, my mistake, not resolve from k8s node labels, but from vm labels.


But we can merge this first, then give another pr to implement it. What do you think? @patrostkowski

patrostkowski: @jwcesign I have pushed the changes for Karpenter config flags.

About the region, cluster name, and project ID — I think those are the minimum inputs we need, and here is why:

Extracting labels from nodes adds unnecessary complexity in my opinion — it turns into a guessing game: trying to infer the cluster name, then the region, and finally the project. That approach might work if we relied on the metadata server, but not when just using an existing node pool.

Right now, my Karpenter deployment runs in a fresh cluster where no node pools exist yet. So there is nothing I can extract info from. By using those three env vars, I explicitly tell the operator which GKE cluster it should manage.

This setup actually works well for two reasons:

1. If we deploy Karpenter inside the cluster, we could use the [metadata server](https://cloud.google.com/kubernetes-engine/docs/how-to/workload-identity) instead to get this info — but that requires Workload Identity, which adds extra setup for users.
2. If we deploy Karpenter to manage an external cluster, setting the region/project/cluster name manually lets us manage clusters in different projects or regions without relying on local metadata.

So overall, I think this change keeps things simple and flexible — but I would be happy to create an issue so we can explore this further in the future.

patrostkowski: for example in crossplane, the ProviderConfig requires the Project ID to be provider before running any resource creation https://docs.crossplane.io/latest/getting-started/provider-gcp/#create-a-providerconfig

jwcesign: Totally agree.

jwcesign (approved):
/lgtm

Issue #16, opened 2025-04-02
Enhancement: Karpenter GCP Provider Development


jwcesign:
### Description

The overall architecture of the code is structured as follows. For reference, you can explore the implementation details in the [Karpenter Provider AlibabaCloud repository](https://github.com/cloudpilot-ai/karpenter-provider-alibabacloud/tree/main/pkg).

### Overview
The **Karpenter core** interacts with the `CloudProvider` interface, which serves as an abstraction layer. The implemented `CloudProvider` then delegates specific operations to various cloud `Providers`, ensuring that the behavior expected by Karpenter core is executed seamlessly.

### CloudProvider Interface
The `CloudProvider` interface defines the following methods:
- [x] **Create**: Calls the respective provider to create a node. @IRONICBo 
- [x] **Delete**: Calls the respective provider to delete a node.
- [x] **Get**: Retrieves a specific node by calling the respective provider.
- [x] **List**: Lists all nodes by invoking the respective provider.
- [x] **GetInstanceTypes**: Fetches available instance types from the provider.
- [x] **IsDrifted**: Checks if a node has drifted (i.e., deviated from its expected state) by querying the provider. @IRONICBo https://github.com/cloudpilot-ai/karpenter-provider-gcp/pull/40
- [x] **RepairPolicies**: Implements logic to repair policies (relatively straightforward to implement). @helen-frank https://github.com/cloudpilot-ai/karpenter-provider-gcp/pull/8
- [x] **Name**: Provides the name of the cloud provider (straightforward to implement). @helen-frank https://github.com/cloudpilot-ai/karpenter-provider-gcp/pull/8
- [x] **GetSupportedNodeClasses**: Retrieves supported node classes (straightforward to implement). @helen-frank  https://github.com/cloudpilot-ai/karpenter-provider-gcp/pull/8

### Providers
The `Providers` module encapsulates cloud-specific functionalities:
- [x] **Node Metadata**: Resolves metadata for GKE nodes. @IRONICBo https://github.com/cloudpilot-ai/karpenter-provider-gcp/pull/34
- [x] **Image Family**: Resolves the self-link of images used by GKE nodes. @jwcesign https://github.com/cloudpilot-ai/karpenter-provider-gcp/pull/15
[...]

PR #20, opened 2025-04-07
feature: implement InstanceType boilerplate code

patrostkowski:
Initial scaffolding for instancetype.Provider including stub methods and controller wiring. Also, adds example NodePool CRD and updates dependencies.

#### What type of PR is this?

feature

none

https://github.com/cloudpilot-ai/karpenter-provider-gcp/issues/16

none

none


patrostkowski:
still working on it, probably I will have some questions I am going to share on Discord

jwcesign:
Hi, @patrostkowski 
Can you help fix the conflicts?

patrostkowski:
ok @jwcesign rebased

examples/crds/default-nodeclass.yaml
```diff
@@ -0,0 +1,22 @@
+apiVersion: karpenter.sh/v1
+kind: NodePool
```
jwcesign, line 2: Maybe put this into `examples` directly?

jwcesign (commented):
Some minor nits, other LGTM.

examples/crds/default-nodeclass.yaml
jwcesign: cc @patrostkowski

patrostkowski: @jwcesign I did this on purpose, as we have a terraform example there too https://github.com/cloudpilot-ai/karpenter-provider-gcp/tree/main/examples

I think the approach is ok, its done similarly in other projects https://github.com/rook/rook/tree/master/deploy/examples

jwcesign: crd stands for custom resource definition. Perhaps we could consider using a different name, such as `config`, for clarity?

patrostkowski: hmm we could move out `examples/crds/*` to simply `examples/` and `examples/deploy` into `deploy/` directory, wdyt? but I would prefer to do it after merging this change to not do two things at the same time

jwcesign: +1

Issue #7, opened 2025-02-01
Method of self-managed node creation


jwcesign:
Karpenter needs to manage the nodes themselves, which means that GKE (Google Kubernetes Engine) should allow self-managed nodes. The following experiment should prove that self-managed nodes are allowed.

please check: https://docs.google.com/document/u/0/d/1FbWWIGn5soHHnRPRxnIxuninW4s64nFuYxjtOJPKXxg/mobilebasic

patrostkowski:
@jwcesign is this demo script already implemented in the controller or is it a pending task?

jwcesign:
Hi, @patrostkowski, 
It's a demo script for pending tasks. If you're interested in implementing it, we welcome your contributions.

PR #28, opened 2025-04-11
feature: implement instance lifecycle for nodeclaims

patrostkowski:
This adds full support for creating, listing, getting and deleting GCP instances through the Karpenter CloudProvider. It also includes label sanitization, proper instance filtering, and a basic CSR auto-approver. Example manifests and CRDs were updated to reflect the new fields.

#### What type of PR is this?

/kind feature
/kind api-change

Implements Instance provider with basic CloudProvider setup

https://github.com/cloudpilot-ai/karpenter-provider-gcp/issues/7

none


patrostkowski:
work in progress, I didnt have a chance to run it against the cluster yet.

pkg/cloudprovider/cloudprovider.go
```diff
@@ -43,32 +51,144 @@ type CloudProvider struct {
...
-	return nil, fmt.Errorf("not implemented")
+	log.FromContext(ctx).Info("listing instances from instance provider")
+	instances, err := c.instanceProvider.List(ctx)
+	if err != nil {
+		log.FromContext(ctx).Error(err, "failed to list instances")
+		return nil, fmt.Errorf("listing instances, %w", err)
+	}
 
+	var nodeClaims []*karpv1.NodeClaim
+	for i := range instances {
```
dm3ch, line 123: Wouldn't it be better here to do like:
```
for _, instance := range instances {
```

And use `instance` instead of `instances[i]` inside the loop

pkg/cloudprovider/cloudprovider.go
```diff
@@ -168,3 +289,60 @@ func (c *CloudProvider) Name() string {
...
+	// Set core labels
+	labels[corev1.LabelTopologyZone] = i.Location
+	labels[karpv1.CapacityTypeLabelKey] = i.CapacityType
+
+	// Add node pool label if present
+	if v, ok := i.Labels[karpv1.NodePoolLabelKey]; ok {
+		labels[karpv1.NodePoolLabelKey] = v
+	}
+
+	nodeClaim.Labels = labels
```
dm3ch, line 318: Wouldn't it make sense to assign it directly to `nodeClaim` structure from `if instanceType != nil`

pkg/cloudprovider/cloudprovider.go
```diff
@@ -168,3 +289,60 @@ func (c *CloudProvider) Name() string {
...
+		return nil, fmt.Errorf("nodeClaim missing NodeClassRef")
+	}
+	nodeClass := &v1alpha1.GCENodeClass{}
+	if err := c.kubeClient.Get(ctx, types.NamespacedName{Name: ref.Name}, nodeClass); err != nil {
+		return nil, err
+	}
+	return nodeClass, nil
+}
+
+func (c *CloudProvider) resolveInstanceTypes(ctx context.Context, nodeClass *v1alpha1.GCENodeClass) ([]*cloudprovider.InstanceType, error) {
```
dm3ch, line 346: Does it make sense to have a separate helper function if it just calls `c.instanceTypeProvider.List` with same arguments and returns results as is?

pkg/utils/utils.go
```diff
@@ -0,0 +1,48 @@
...
+	}
+	for key, req := range instanceType.Requirements {
+		if req.Len() == 1 {
+			labels[key] = req.Values()[0]
+		}
+	}
+	return labels
+}
+
+func SanitizeGCELabelValue(s string) string {
```
dm3ch, line 42: Wouldn't it be better not to sanitise values on creation, but to validate them on CRD / etc level before it faces GCP API, or it's done in a similar way in other providers?

By the way, I'm not so good in regexps but wouldn't the ReplaceAllString replace all alphanumeric sequences with just a `-` symbol? And if so is it intentional?

dm3ch: Just a quick question — does AWS Karpenter perform any kind of label sanitization?
I’m asking to understand whether this is a GCP-specific concern or more of a precautionary measure.

From what I’ve found so far, Karpenter only checks that custom tags don’t overlap with system ones:
https://github.com/aws/karpenter-provider-aws/blob/main/pkg/apis/v1/labels.go#L71

patrostkowski:
hey @IRONICBo I am back from holidays, I have seen that you pushed the changes to this PR but it looks like a rebase against main branch, is it correct?

IRONICBo:
> hey @IRONICBo I am back from holidays, I had seen that you pushed the changes to this PR but it looks like a rebase against main branch, is it correct?

Yes, I'll push the remaining commits today. @patrostkowski

pkg/cloudprovider/cloudprovider.go
IRONICBo: I agree with it, and I have update this operation.

examples/nodeclaim/default-nodeclaim.yaml
```diff
@@ -0,0 +1,24 @@
...
+    - key: "karpenter.sh/capacity-type"
+      operator: In
+      values: ["on-demand"]
+    - key: "kubernetes.io/arch"
+      operator: In
+      values: ["amd64"]
+    - key: "topology.kubernetes.io/zone"
+      operator: In
+      values: ["us-central1-c", "us-central1-a", "us-central1-f", "us-central1-b"]
+    - key: "node.kubernetes.io/instance-type"
```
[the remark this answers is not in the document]
IRONICBo: Thank you, in the Karpenter node claim use will indeed support the instance family label here.

The demo YAML may be unnecessary here, nodeclaim will be created automatically by karpenter, for users only nodepool and nodeclass need to be created.

So, for the sake of simplicity, I remove the nodeclaim example yaml here.

pkg/apis/v1alpha1/gcenodeclass.go
```diff
@@ -136,6 +147,21 @@ type KubeletConfiguration struct {
...
+	Name string `json:"name,omitempty"`
+	// Tags is a map of labels used to select subnetworks
+	Tags map[string]string `json:"tags,omitempty"`
+}
+
+type DiskSpec struct {
+	// SizeGiB is the size of the boot disk
+	// +kubebuilder:validation:Minimum=10
+	SizeGiB int32 `json:"sizeGiB"`
+	// Type of the disk (e.g., pd-standard, pd-balanced, pd-ssd)
```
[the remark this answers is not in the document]
IRONICBo: I have checked the current disk types, and the supported types are as follows.

```bash
gcloud compute disk-types list --format="table[no-heading](name)" | sort | uniq
hyperdisk-balanced
hyperdisk-balanced-high-availability
hyperdisk-extreme
hyperdisk-ml
hyperdisk-throughput
local-ssd
pd-balanced
pd-extreme
pd-ssd
pd-standard
```

Besides, using enum is a better way to check CRD type here.

examples/nodeclaim/default-nodeclaim.yaml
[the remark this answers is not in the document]
dm3ch: Make sense.

But we already having a nodepool / nodeclass examples that lacks the family label, should we add it there as part of this PR?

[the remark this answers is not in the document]
IRONICBo: OK, I will add it later.

pkg/cloudprovider/cloudprovider.go
IRONICBo: Thank you for your suggestion, but when instance type is nil, we can still set the karpenter label for nodeclaim to facilitate subsequent processing by karpenter, so we can keep the current implementation.

pkg/cloudprovider/cloudprovider.go
IRONICBo: Thank you for your suggestion, but when instance type is nil, we can still set the karpenter label for nodeclaim to facilitate subsequent processing by karpenter, so we can keep the current implementation.

pkg/cloudprovider/cloudprovider.go
IRONICBo: Yes you are correct, but like other karpenter providers, we can provide some filtering methods in this method to enhance the available results after List.

https://github.com/cloudpilot-ai/karpenter-provider-alibabacloud/blob/a8e99d9f292d640ee0bce334d24240cb15501e7f/pkg/cloudprovider/cloudprovider.go#L341

pkg/apis/v1alpha1/gcenodeclass.go
```diff
@@ -30,9 +30,9 @@ type GCENodeClassSpec struct {
 	// ServiceAccount is the GCP IAM service account email to assign to the instance
 	// +kubebuilder:validation:Pattern=`^[^@]+@(developer\.gserviceaccount\.com|[^@]+\.iam\.gserviceaccount\.com)$`
 	ServiceAccount string `json:"serviceAccount"`
-	// Disk defines the boot disk to attach to the instance.
+	// Disk defines the boot disk to attach to the provisioned instance.
 	// +optional
-	Disk *DiskSpec `json:"disk,omitempty"`
+	SystemDisk *SystemDisk `json:"disk,omitempty"`
```
patrostkowski, line 35: @IRONICBo I was initially mistaken in saying this field sets the boot disk size, because it actually represents the list of disks to attach to the instance.

That said, I’m a bit confused about the name change for the Disks field, especially since the Instance type in the [official docs](https://pkg.go.dev/google.golang.org/api/compute/v1#Instance) doesn't have such a field explicitly named like that.

Maybe we should revert the name back to `Disks` and update the description to:
`Disks define the list of disks to attach to the provisioned instance.`

pkg/utils/utils.go
IRONICBo: Make sense, the sanitization operation here may not be necessary. Currently, all the values ​​passed into the label are const and legal values, so the sanitization operation may not be necessary.

cc @jwcesign @dm3ch

pkg/apis/v1alpha1/gcenodeclass.go
IRONICBo: Make sense, but we want to mark the disk as a system disk or a normal data disk when creating the node and actually running the node, and align it with our other provider.

https://github.com/cloudpilot-ai/karpenter-provider-alibabacloud/blob/a8e99d9f292d640ee0bce334d24240cb15501e7f/pkg/apis/v1alpha1/ecsnodeclass.go#L76

patrostkowski: It works a bit differently in GCP, as the VM is booted from an image (with a configurable size) rather than an actual disk. Therefore, I think it's more appropriate to rename the field to `Disks`.

image

dm3ch: What about disks used for /var/lib/docker and etc? Is it live on additional disk by default or it's part of partition created by system disk?

pkg/apis/v1alpha1/gcenodeclass.go
```diff
@@ -136,6 +147,30 @@ type KubeletConfiguration struct {
...
+// If multiple fields are used for selection, the requirements are ANDed.
+type SubnetworkSelectorTerm struct {
+	// Name of the subnetwork
+	// +optional
+	Name string `json:"name,omitempty"`
+	// Tags is a map of labels used to select subnetworks
+	// +kubebuilder:validation:XValidation:message="empty tag keys or values aren't supported",rule="self.all(k, k != '' && self[k] != '')"
+	// +kubebuilder:validation:MaxProperties:=20
+	// +optional
+	Tags map[string]string `json:"tags,omitempty"`
```
jwcesign, line 160: Subnet hasn't tags.
```json
{
  "autoCreateSubnetworks": true,
  "gatewayAddress": "10.128.0.1",
  "id": "7480249233495977737",
  "ipCidrRange": "10.128.0.0/20",
  "name": "default",
  "networkUrl": "projects/cloudpilot-ai-439712/global/networks/default",
  "privateIpGoogleAccess": true,
  "purpose": "PURPOSE_PRIVATE",
  "regionUrl": "projects/cloudpilot-ai-439712/regions/us-central1",
  "secondaryIpRanges": [
    {
      "ipCidrRange": "10.4.0.0/14",
      "name": "gke-cloudpilot-production-pods-1058d762",
      "reservedInternalRangeUrl": "projects/cloudpilot-ai-439712/locations/global/internalRanges/gke-cloudpilot-production-pods-1058d762"
    },
    {
      "ipCidrRange": "10.64.0.0/14",
      "name": "gke-cluster-1-test-pods-cff58a70",
      "reservedInternalRangeUrl": "projects/cloudpilot-ai-439712/locations/global/internalRanges/gke-cluster-1-test-pods-cff58a70"
    },
    {
      "ipCidrRange": "10.56.0.0/14",
      "name": "gke-helen-c1-spot-pods-3e5f97cd",
      "reservedInternalRangeUrl": "projects/cloudpilot-ai-439712/locations/global/internalRanges/gke-helen-c1-spot-pods-3e5f97cd"
    }
  ],
  "selfLink": "projects/cloudpilot-ai-439712/regions/us-central1/subnetworks/default",
  "stackType": "IPV4_ONLY"
}
```

pkg/apis/v1alpha1/gcenodeclass.go
```diff
@@ -44,6 +48,10 @@ type GCENodeClassSpec struct {
 	// +kubebuilder:validation:XValidation:message="evictionSoftGracePeriod OwnerKey does not have a matching evictionSoft",rule="has(self.evictionSoftGracePeriod) ? self.evictionSoftGracePeriod.all(e, (e in self.evictionSoft)):true"
 	// +optional
 	KubeletConfiguration *KubeletConfiguration `json:"kubeletConfiguration,omitempty"`
+	// SubnetworkSelectorTerms is a list of subnetwork selector terms. The terms are ANDed.
+	// +kubebuilder:validation:MinItems:=1
+	// +kubebuilder:validation:MaxItems:=30
+	SubnetworkSelectorTerms []SubnetworkSelectorTerm `json:"subnetworkSelectorTerms"`
```
jwcesign, line 54: I think we do not need this filed.
image

The MVP, we can resolve the vpc/subnet from gke cluster information

pkg/controllers/csr/csr.go
```diff
@@ -0,0 +1,92 @@
...
+	for _, csr := range csrList.Items {
+		if isApprovedOrDenied(&csr) {
+			continue
+		}
+
+		if csr.Spec.SignerName != KubeletClientSignerName {
+			continue
+		}
+
+		if !reflect.DeepEqual(csr.Spec.Usages, []certv1.KeyUsage{
```
jwcesign, line 51: Does this only approve the bootstrap CSR?

pkg/providers/instance/instance.go
```diff
@@ -34,36 +44,326 @@ type Provider interface {
...
+}
+
+func (p *DefaultProvider) findTemplateForAlias(ctx context.Context, alias string) (*compute.InstanceTemplate, error) {
+	if alias == "" {
+		return nil, fmt.Errorf("alias not specified in ImageSelectorTerm")
+	}
+
+	var expectedLabelValue string
+	switch alias {
+	case "ContainerOptimizedOS":
```
jwcesign, line 135: Defile this as const value

jwcesign (commented):
Generally lgtm. Only few comments

pkg/apis/v1alpha1/gcenodeclass.go
patrostkowski: @dm3ch disks mounted via `Disks` in the Instance struct will be available on the host under `/dev` directory but will remain unmounted. The root is actually mounted in the VM image itself.

@jwcesign I will update the code to follow the console then.

pkg/apis/v1alpha1/gcenodeclass.go
patrostkowski: yup, lets go this route. I was mistaken by the network tags attached to the nodes. Subnets can be defined only in multi-network clusters, lets not focus on that right now.

image

pkg/controllers/csr/csr.go
patrostkowski: Yep, this controller is meant to approve only bootstrap CSRs. It looks for ones with the signer name `kubernetes.io/kube-apiserver-client-kubelet` and a specific set of usages: digital signature, key encipherment, and client auth. Any other CSRs are skipped.

pkg/providers/instance/instance.go
patrostkowski: absolutely!

pkg/apis/v1alpha1/gcenodeclass.go
patrostkowski: Indeed, its the hosts that have network tags attached, let me adjust that

image

pkg/apis/v1alpha1/gcenodeclass.go
patrostkowski: @jwcesign  I tried look up if we can define the disks in some reasonable way, but I think it would require much more work than I expected and I am afraid that this PR would become very messy. 

Saying that, I have opened https://github.com/cloudpilot-ai/karpenter-provider-gcp/issues/36 so we can look it up later. I will update with links and code snippets once this PR is merged.

examples/nodeclaim/default-nodeclaim.yaml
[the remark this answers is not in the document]
patrostkowski: @IRONICBo I think once this is merged, we could try to make it actually deduct what type of node it should create, because currently I create just a static e2-standard-2 machine.

jwcesign (commented):
/lgtm

cc @dm3ch for final checking.

pkg/utils/utils.go
patrostkowski: sure, we could comment on it. also, we could even just create constant variables with sanitized labels, so we dont waste CPU cycles on computing those values.

PR #29, opened 2025-04-11
doc: add Compute Admin GCP Role to SA requirement

froblesmartin:
#### What type of PR is this?

/kind documentation

Add a new GCP Role required for Karpenter to run.

None

We should split the required roles to create the GKE cluster and the roles required for Karpenter to run. If you agree, let me know and I can change this PR to do that 👍 .

```release-note
doc: add Compute Admin GCP Role to SA requirement
```

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -28,6 +28,7 @@ This guides breaks down into two parts
 
 3. Next go ahead and create the service account in the project and assign the following role to the service account
 
+    - Compute Admin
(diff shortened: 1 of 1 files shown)


jwcesign:
> We should split the required roles to create the GKE cluster and the roles required for Karpenter to run. If you agree, let me know and I can change this PR to do that 👍 .

It would be better. Thanks @froblesmartin

jwcesign:
We can create an additional PR to do that.

/lgtm

PR #32, opened 2025-04-21
Fix deadlock on instancetype provider List() invokation

dm3ch:
List was locking mutex in the beging of invokation and then called getInstanceTypes which also tried to acquire the lock. Lock acquirition was dropped from List() function cause it doesn't directly interacts with the cache, so it doesn't need it.

#### What type of PR is this?
/kind bug


patrostkowski:
> I also wanted to write some tests to avoid breaking the provider, but found out that it's tricky cause it uses GCP compute provider, which needs to be mocked to test anything. :(

I thought the same. We should find a way to mock the API and write some basic tests at some point.

PR #42, opened 2025-05-21
feat: create instance tags op.

IRONICBo:
#### What type of PR is this?

/kind feature

This PR supports creating tags for Karpenter provider, so controller can update instance tags and find the instance managed by controller.

Fixes #

```release-note
Support create tags op for karpenter.
```

### DIFF
--- a/pkg/providers/instance/instance.go
+++ b/pkg/providers/instance/instance.go
@@ -49,6 +49,7 @@ type Provider interface {
 }
 
 type DefaultProvider struct {
+	// In current implementation, instanceID == InstanceName
 	instanceCache *cache.Cache
 
 	region         string
@@ -210,6 +211,10 @@ func (p *DefaultProvider) Get(ctx context.Context, providerID string) (*Instance
 		return nil, fmt.Errorf("parsing provider ID: %w", err)
 	}
 
+	if instance, ok := p.instanceCache.Get(instanceName); ok {
+		return instance.(*Instance), nil
+	}
+
 	log := log.FromContext(ctx)
 	log.Info("Fetching instance", "project", project, "zone", zone, "instance", instanceName)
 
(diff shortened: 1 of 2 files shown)


pkg/providers/instance/instance.go
```diff
@@ -301,6 +309,9 @@ func (p *DefaultProvider) List(ctx context.Context) ([]*Instance, error) {
 	}
 
 	log.FromContext(ctx).Info("finished listing GCP instances", "total", len(instances))
+
+	p.syncAllInstances(instances)
```
patrostkowski, line 313: maybe each instance could be cached in the loop above instead? we could handle both operations at the same time

pkg/providers/instance/instance.go
```diff
@@ -340,8 +351,44 @@ func (p *DefaultProvider) Delete(ctx context.Context, providerID string) error {
...
-	// TODO: Implement me
-	return nil
+	projectID, zone, instanceName, err := parseGCEProviderID(providerID)
+	if err != nil {
+		return fmt.Errorf("failed to parse provider ID: %w", err)
+	}
+
+	instance, err := p.computeService.Instances.Get(projectID, zone, instanceName).Context(ctx).Do()
+	if err != nil {
+		if gErr, ok := err.(*googleapi.Error); ok && gErr.Code == 404 {
```
patrostkowski, line 361: I think this requires something like:
```
func isInstanceNotFoundError(err error) bool {
	var apiErr *googleapi.Error
	if errors.As(err, &apiErr) {
		return apiErr.Code == 404
	}
	return false
}
```

pkg/providers/instance/instance.go
IRONICBo: Make sense, I will check it.

pkg/providers/instance/instance.go
IRONICBo: OK, I will update it.

IRONICBo: @patrostkowski I have changed to `isInstanceNotFoundError`.

patrostkowski: ok thats even better!

patrostkowski:
/lgtm

@jwcesign I think you might also want to take a look at this

jwcesign (commented):
/approve

PR #43, opened 2025-05-22
fix: support get instance type with frequency detecting.

IRONICBo:
#### What type of PR is this?

/kind feature

Add frequency detection in 

Fixes #

```release-note
support get instance type with frequency detecting.
```

### DIFF
--- a/pkg/cloudprovider/cloudprovider.go
+++ b/pkg/cloudprovider/cloudprovider.go
@@ -19,6 +19,7 @@ import (
 	stderrors "errors"
 	"fmt"
 	"net/http"
+	"strconv"
 	"time"
 
 	"github.com/awslabs/operatorpkg/status"
@@ -42,7 +43,10 @@ import (
 	"github.com/cloudpilot-ai/karpenter-provider-gcp/pkg/utils"
 )
 
-const CloudProviderName = "gcp"
+const (
+	CloudProviderName                       = "gcp"
+	CloudPilotDiversityInstanceTypeLabelKey = "cloudpilot.ai/diversity-instancetype"
+)
 
 var _ cloudprovider.CloudProvider = (*CloudProvider)(nil)
 
@@ -155,6 +159,7 @@ func (c *CloudProvider) resolveInstanceTypeFromInstance(ctx context.Context, ins
 	if instanceType == nil {
 		return nil, fmt.Errorf("instance type %s not found in offerings", instance.Type)
 	}
+
 	return instanceType, nil
 }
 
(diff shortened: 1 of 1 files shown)


pkg/cloudprovider/cloudprovider.go
```diff
@@ -200,17 +205,119 @@ func (c *CloudProvider) GetInstanceTypes(ctx context.Context, nodePool *karpv1.N
...
 	}
-	// TODO, break this coupling
 	instanceTypes, err := c.instanceTypeProvider.List(ctx, nodeClass)
 	if err != nil {
+		log.FromContext(ctx).Error(err, "listing instance types")
 		return nil, err
 	}
+
+	var diversityInstanceType bool
+	if v, ok := nodePool.Labels[CloudPilotDiversityInstanceTypeLabelKey]; ok {
```
patrostkowski, line 221: I think we could do it like `diversityInstanceType, _ := strconv.ParseBool(nodePool.Labels[CloudPilotDiversityInstanceTypeLabelKey])
` without `ok` check

pkg/cloudprovider/cloudprovider.go
```diff
@@ -200,17 +205,119 @@ func (c *CloudProvider) GetInstanceTypes(ctx context.Context, nodePool *karpv1.N
...
+
+	return filteredInstanceTypes, nil
+}
+
+// filterInstanceTypesByMaxSpotInstanceType disables Spot offerings on the most used Spot instance type.
+// It scans all NodeClaims labeled as Spot and determines which instance type is used most frequently (with tiebreaker on creation time).
+// Then it disables all Spot offerings for that specific instance type.
+//
+//nolint:gocyclo
+func (c *CloudProvider) filterInstanceTypesByMaxSpotInstanceType(
```
patrostkowski, line 242: I think this method could be split further into smaller methods, like: `mostUsedSpotInstanceType` and so on, just to extend testability and approach it more atomic.

pkg/cloudprovider/cloudprovider.go
```diff
@@ -200,17 +205,119 @@ func (c *CloudProvider) GetInstanceTypes(ctx context.Context, nodePool *karpv1.N
...
+
+		instanceType, ok := nc.Labels[corev1.LabelInstanceTypeStable]
+		if !ok || instanceType == "" {
+			continue
+		}
+
+		if stat, exists := existingSpotInstanceTypeM[instanceType]; exists {
+			stat.count++
+			// Add creation time for tie-breaking
+			stat.createTimeNum += nc.CreationTimestamp.Unix()
```
patrostkowski, line 281: I think this would benefit from struct to calculate and usage of https://pkg.go.dev/time#Time.Before

pkg/cloudprovider/cloudprovider.go
```diff
@@ -200,17 +205,119 @@ func (c *CloudProvider) GetInstanceTypes(ctx context.Context, nodePool *karpv1.N
...
+	// Fetch all NodeClaims to analyze current Spot usage
+	nodeClaimList := &karpv1.NodeClaimList{}
+	if err := c.kubeClient.List(ctx, nodeClaimList); err != nil {
+		return nil, fmt.Errorf("listing nodeclaims, %w", err)
+	}
+
+	existingSpotInstanceTypeM := make(map[string]*instanceTypeCount)
+
+	// Aggregate count and creation time of Spot NodeClaims per instance type
+	for i := range nodeClaimList.Items {
```
patrostkowski, line 266: I think we could exit earlier here:

```
// Early check for both required labels
capacityType := labels[karpv1.CapacityTypeLabelKey]
instanceType := labels[corev1.LabelInstanceTypeStable]
if capacityType != karpv1.CapacityTypeSpot || instanceType == "" {
    continue
}
```

patrostkowski (changes requested):
requires some polishing, but approach is very good

pkg/cloudprovider/cloudprovider.go
IRONICBo: Make sense, I think it is better.

pkg/cloudprovider/cloudprovider.go
IRONICBo: OK, this function is a bit large, better to split it.

pkg/cloudprovider/cloudprovider.go
IRONICBo: Thanks!

pkg/cloudprovider/cloudprovider.go
IRONICBo: Make sense, I will update it.

pkg/cloudprovider/cloudprovider.go
```diff
@@ -42,7 +43,10 @@ import (
 	"github.com/cloudpilot-ai/karpenter-provider-gcp/pkg/utils"
 )
 
-const CloudProviderName = "gcp"
+const (
+	CloudProviderName                       = "gcp"
+	CloudPilotDiversityInstanceTypeLabelKey = "cloudpilot.ai/diversity-instancetype"
```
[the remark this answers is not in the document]
IRONICBo: Unused feature, I have removed them, thanks.

IRONICBo:
Sorry this PR is unused in current stage, it is better to close it. We want need to resolve it later.

cc @patrostkowski

PR #44, opened 2025-05-23
feat: add instance type and liveness probe.

IRONICBo:
#### What type of PR is this?

Support to query instance type with liveness probe and 

/kind feature

Check the instance type and support the liveness probe function.

Fixes #

```release-note
Add instance type and liveness probe.
```

### DIFF
--- a/pkg/providers/instancetype/instancetype.go
+++ b/pkg/providers/instancetype/instancetype.go
@@ -1,4 +1,6 @@
 /*
+Copyright 2025 The CloudPilot AI Authors.
+
 Licensed under the Apache License, Version 2.0 (the "License");
 you may not use this file except in compliance with the License.
 You may obtain a copy of the License at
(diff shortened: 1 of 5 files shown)


pkg/providers/instancetype/instancetype.go
```diff
@@ -1,4 +1,6 @@
 /*
+Copyright 2024 The CloudPilot AI Authors.
```
jwcesign, line 2: 2025

pkg/providers/instancetype/types.go
```diff
@@ -0,0 +1,124 @@
+/*
+Copyright 2024 The CloudPilot AI Authors.
```
jwcesign, line 2: 2025

IRONICBo: Got.

PR #47, opened 2025-06-04
fix: update region operation client to zone operation client.

IRONICBo:
#### What type of PR is this?

/kind bug

Change operation from region operator to zone operator.

Fixes #

```release-note
fix: update region operation client to zone operation client.
```

cc @patrostkowski

### DIFF
--- a/pkg/controllers/interruption/controller.go
+++ b/pkg/controllers/interruption/controller.go
@@ -55,14 +55,14 @@ type Controller struct {
 
 	unavailableOfferingsCache *cache.UnavailableOfferings
 	metadataClient            *metadata.Client
-	operationClient           *computev1.RegionOperationsClient
+	zoneOperationClient       *computev1.ZoneOperationsClient
 	credential                auth.Credential
 }
 
 func NewController(kubeClient client.Client, recorder events.Recorder,
 	unavailableOfferingsCache *cache.UnavailableOfferings,
 	metadataClient *metadata.Client,
-	operationClient *computev1.RegionOperationsClient,
+	zoneOperationClient *computev1.ZoneOperationsClient,
 	credential auth.Credential) *Controller {
 	return &Controller{
 		kubeClient: kubeClient,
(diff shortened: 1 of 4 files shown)


pkg/controllers/interruption/controller.go
```diff
@@ -55,71 +55,108 @@ type Controller struct {
...
+	if err := kubeClient.List(ctx, nodeList, &client.ListOptions{}); err != nil {
+		return nil, fmt.Errorf("listing nodes: %w", err)
+	}
+
+	zoneSet := make(map[string]struct{})
+	for _, node := range nodeList.Items {
+		zone := node.Labels["topology.kubernetes.io/zone"]
+		if zone == "" {
+			// fallback to legacy label
+			zone = node.Labels["failure-domain.beta.kubernetes.io/zone"]
```
jwcesign, line 90: Check this also

pkg/controllers/interruption/controller.go
```diff
@@ -55,71 +55,108 @@ type Controller struct {
...
-			log.FromContext(ctx).Error(err, "listing operations")
-			break
+// getZonesFromNodes lists all zones by inspecting node labels
+func getZonesFromNodes(ctx context.Context, kubeClient client.Client) ([]string, error) {
+	nodeList := &corev1.NodeList{}
+	if err := kubeClient.List(ctx, nodeList, &client.ListOptions{}); err != nil {
+		return nil, fmt.Errorf("listing nodes: %w", err)
+	}
+
+	zoneSet := make(map[string]struct{})
```
jwcesign, line 85: Maybe use set.String?

jwcesign (commented):
Other lgtm

pkg/controllers/interruption/controller.go
IRONICBo: Ditto, changed to well-known labels in apicore.

pkg/controllers/interruption/controller.go
IRONICBo: I have changed to `zoneSet := sets.Set[string]{}` :)

pkg/controllers/interruption/controller.go
```diff
@@ -55,71 +55,108 @@ type Controller struct {
...
+	for _, node := range nodeList.Items {
+		zone := node.Labels["topology.kubernetes.io/zone"]
```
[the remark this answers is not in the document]
patrostkowski: yep, using const values is the best here.

PR #49, opened 2025-06-09
fix: choose node template for current cluster in case of mixed cluster csr.

IRONICBo:
#### What type of PR is this?

/kind bug

Current `Create` function will choose first node template, and this template is shared in current workspace. Template with wrong metadata will cause init error.

Fixes #

```release-note
fix: choose node template for current cluster in case of mixed cluster csr.
```

cc @patrostkowski


jwcesign (approved):
/approve

Thanks @IRONICBo

PR #50, opened 2025-06-10
fix: remove taint for ready nodes.

IRONICBo:
#### What type of PR is this?

/kind bug

Nodeclaim controller in karpenter will check `StatusConditions` in nodeclaim, contains `UnregisteredNoExecuteTaint`, we need to remove this taint when node is ready.

Fixes #

```release-note
fix: remove taint for ready nodes.
```


jwcesign:
This taint is removed by Karpenter core, why do we need to implement another controller to do this?

IRONICBo:
> This taint is removed by Karpenter core, why do we need to implement another controller to do this?

An error occurred in GCP, so I want to manually delete it.

patrostkowski:
> > > > This taint is removed by Karpenter core, why do we need to implement another controller to do this?
> > > 
> > > 
> > > An error occurred in GCP, so I want to manually delete it.
> > 
> > 
> > What kind of error, can you paste it here?
> 
> UnregisteredTaintNotFound is illegal here.
> 
> ```shell
> message: Invariant violated, karpenter.sh/unregistered taint must be present on
>   Karpenter-managed nodes
> observedGeneration: 1
> reason: UnregisteredTaintNotFound
> status: "False"
> type: Registered
> 
> message: Registered=False, Initialized=Unknown
> observedGeneration: 1
> reason: UnhealthyDependents
> status: "False"
> type: Ready
> ```

not sure if this is correct approach to imlement controller for this. could you share more logs what actually happened?

jwcesign:
> message: Invariant violated, karpenter.sh/unregistered taint must be present on
  Karpenter-managed nodes


When ceate nodes, you should set the startup taints of kubelet. Please try to fix in this way.

One demonstration: https://github.com/cloudpilot-ai/karpenter-provider-alibabacloud/blob/776d09e6fdf7ea3728c1e7b838eb3998f3e36e9b/pkg/providers/instance/instance.go#L689

IRONICBo:
> > message: Invariant violated, karpenter.sh/unregistered taint must be present on
> > Karpenter-managed nodes
> 
> When ceate nodes, you should set the startup taints of kubelet. Please try to fix in this way.
> 
> One demonstration: https://github.com/cloudpilot-ai/karpenter-provider-alibabacloud/blob/776d09e6fdf7ea3728c1e7b838eb3998f3e36e9b/pkg/providers/instance/instance.go#L689

Thanks, I accidentally neglected this, I will add it.

IRONICBo:
> > > > > This taint is removed by Karpenter core, why do we need to implement another controller to do this?
> > > > 
> > > > 
> > > > An error occurred in GCP, so I want to manually delete it.
> > > 
> > > 
> > > What kind of error, can you paste it here?
> > 
> > 
> > UnregisteredTaintNotFound is illegal here.
> > ```shell
> > message: Invariant violated, karpenter.sh/unregistered taint must be present on
> >   Karpenter-managed nodes
> > observedGeneration: 1
> > reason: UnregisteredTaintNotFound
> > status: "False"
> > type: Registered
> > 
> > message: Registered=False, Initialized=Unknown
> > observedGeneration: 1
> > reason: UnhealthyDependents
> > status: "False"
> > type: Ready
> > ```
> 
> not sure if this is correct approach to imlement controller for this. could you share more logs what actually happened?

This is the relevant log, but this controller does not seem to solve the problem correctly. I will modify it.

PR #51, opened 2025-06-10
chore: add helm for karpenter gcp provider.

IRONICBo:
#### What type of PR is this?

/kind feature

Add helm config for gcp provider installation.

Fixes #

```release-note
chore: add helm for karpenter gcp provider.
```

### DIFF
--- a/charts/karpenter/templates/csr.yaml
+++ b/charts/karpenter/templates/csr.yaml
@@ -1,0 +1,12 @@
+apiVersion: rbac.authorization.k8s.io/v1
+kind: ClusterRoleBinding
+metadata:
+  name: karpenter-cert-manager-binding
+subjects:
+- kind: ServiceAccount
+  name: gcp-release-karpenter
+  namespace: default
+roleRef:
+  kind: ClusterRole
+  name: cluster-admin
+  apiGroup: rbac.authorization.k8s.io
--- a/charts/karpenter/templates/gcenodeclass-patch-clusterrole.yaml
+++ b/charts/karpenter/templates/gcenodeclass-patch-clusterrole.yaml
@@ -1,0 +1,43 @@
+apiVersion: rbac.authorization.k8s.io/v1
+kind: ClusterRole
+metadata:
+  name: karpenter-gcenodeclass-patch
+rules:
+  - apiGroups: ["karpenter.k8s.gcp"]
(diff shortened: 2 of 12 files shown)


charts/karpenter/templates/secret.yaml
```diff
@@ -0,0 +1,20 @@
+apiVersion: v1
```
patrostkowski, line 1: I think providing the secret should be the administrator task, so I would opt in having a secret mount mechanism via values.

Another issue is that in GKE users most probably will use https://cloud.google.com/kubernetes-engine/docs/how-to/workload-identity

IRONICBo: Yes, I agree with your point of view. The secret here is mainly for the convenience of development and debugging, and is an optional feature.

lewandowskim1988: Maybe way to replace it and make `helm chart` more versatile is to add possibility to deploy some extra objects like in [this](https://github.com/bitnami/charts/blob/main/bitnami/postgresql/templates/extra-list.yaml#L6) chart? In that case users will not be limited to WI and there will be easy way of adding extra manifests if needed.

jwcesign: it's great idea! @lewandowskim1988 

This PR need more works, would you mind helping about this?

charts/karpenter/templates/gcenodeclass-patch-clusterrole.yaml
lewandowskim1988, line 1: The same as with `csr.yaml` file, what is the purpose of this configuration?

charts/karpenter/templates/csr.yaml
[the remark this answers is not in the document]
jwcesign: cc @IRONICBo Would you mind  reminding us about this?

charts/karpenter/templates/gcenodeclass-patch-clusterrole.yaml
IRONICBo: You need this csr in order to provide the karpenter permission control node, this can be merged with the cr file

jwcesign:
Supported now:
https://github.com/cloudpilot-ai/karpenter-provider-gcp/tree/main/charts

PR #57, opened 2025-06-24
fix: update instance capacity with right pod size.

IRONICBo:
#### What type of PR is this?

/kind bug

Karpenter requirement will fail in mocked scheduling, because the pod is always 0 and all instance type are cleared.

Fixes #

```release-note
fix: update instance capacity with right pod size.
```


jwcesign (approved):
/approve

PR #106, opened 2025-08-07
fix(crd): correct schema for systemDisk category validation

cr0wnf0x:
#### What type of PR is this?
/kind bug

The enum for `systemDisk.categories` was incorrectly placed at the array level instead of the item level. This prevented the validation from being correctly applied to the items within the array.

This PR fixes a bug in the GCENodeClass CRD where the validation for spec.systemDisk.categories was not being applied correctly.

The enum constraint was mistakenly applied to the categories array field itself, rather than to the individual string items within the array. This meant that no actual validation was occurring for the disk types provided by the user.

This change moves the enum to the items property, which is the correct location for validating elements of an array according to the OpenAPI v3 schema specification. This ensures that any value supplied in the categories array must be one of the officially supported GCP disk types.

### DIFF
--- a/charts/karpenter/crds/karpenter.k8s.gcp_gcenodeclasses.yaml
+++ b/charts/karpenter/crds/karpenter.k8s.gcp_gcenodeclasses.yaml
@@ -64,6 +64,7 @@ spec:
                     description: The category of the system disk (e.g., pd-standard,
                       pd-balanced, pd-ssd, pd-extreme).
                     items:
+                      description: DiskCategory represents a disk category type
                       enum:
                       - hyperdisk-balanced
                       - hyperdisk-balanced-high-availability
@@ -76,6 +77,7 @@ spec:
                       - pd-ssd
                       - pd-standard
                       type: string
+                    maxItems: 10
                     type: array
                   sizeGiB:
                     description: 'SizeGiB is the size of the system disk. Unit: GiB'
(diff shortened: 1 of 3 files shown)


jwcesign:
Hi, I think we need to update here: https://github.com/cloudpilot-ai/karpenter-provider-gcp/blob/13017f237809788852abeb6abade3bb764b5a869/pkg/apis/v1alpha1/gcenodeclass.go#L165

To:
```go
type Disk struct {
	// SizeGiB is the size of the system disk. Unit: GiB
	// +kubebuilder:validation:XValidation:message="size invalid",rule="self >= 10"
	// +optional
	SizeGiB int32 `json:"sizeGiB"`
	// The category of the system disk (e.g., pd-standard, pd-balanced, pd-ssd, pd-extreme).
	// +kubebuilder:validation:MaxItems=10
	// +optional
	Categories []DiskCategory `json:"categories,omitempty"`
	// Indicates that this is a boot disk
	// +optional
	Boot bool `json:"boot"`
}

// DiskCategory represents a disk category type
// +kubebuilder:validation:Enum=hyperdisk-balanced;hyperdisk-balanced-high-availability;hyperdisk-extreme;hyperdisk-ml;hyperdisk-throughput;local-ssd;pd-balanced;pd-extreme;pd-ssd;pd-standard
type DiskCategory string
```

And then run `bash hack/update-generated.sh`

cr0wnf0x:
Thanks, missed that the CRD is generated. Will prepare the change and push later today or tomorrow.

jwcesign:
/lgtm

Thanks @cr0wnf0x 
May I know are you running this in your GKE clusters now?

cr0wnf0x:
@jwcesign yes, started implementing it on Monday to be used by our Build Agents.

PR #134, opened 2025-10-02
Fix max pods per node consistency

### LINKED ISSUE #133: "consistency error" while using max pods setting
**What happened**:
When setting maxPods in kubelet config, a consistency error appears due to default max pods setting in nodeclaim

```
{
    "controllerGroup": "karpenter.sh",
    "level": "ERROR",
    "commit": "a6e9e95-dirty",
    "namespace": "",
    "controllerKind": "NodeClaim",
    "error": "expected 110 of resource pods, but found 64 (58.2% of expected)",
    "NodeClaim": {
      "name": "default-spot-ltnsb"
    },
    "logger": "controller",
    "Node": {
      "name": "karpenter-default-spot-ltnsb"
    },
    "name": "default-spot-ltnsb",
    "message": "consistency error",
    "controller": "nodeclaim.consistency",
    "reconcileID": "9b6c3574-bd1a-4a88-b40c-5f429bf7b06f"
}
```

No error

Set max pods in kubelet config

**Environment**:

- Karpenter-provider-gcp version (use `git describe --tags --dirty --always`): a6e9e95-dirty
- GKE version: v1.32.6-gke.1060000
- Others:

sthomson-wyn:
- centralize retrieving max pods per node
- set pod capacity in node claim based on max pods configuration
- default 110 pods per node

#### What type of PR is this?
/kind bug

Fixes consistency error

Fixes #133 

This sets the default max pods per node to the kubelet default of 110, rather than using the cluster's default configuration
```release-note

```

### DIFF
--- a/pkg/providers/instance/instance.go
+++ b/pkg/providers/instance/instance.go
@@ -440,12 +440,8 @@ func (p *DefaultProvider) buildInstance(nodeClaim *karpv1.NodeClaim, nodeClass *
 
 // nolint:gocyclo
 func (p *DefaultProvider) setupNetworkInterfaces(template *compute.InstanceTemplate, nodeClass *v1alpha1.GCENodeClass) []*compute.NetworkInterface {
-	if nodeClass.Spec.KubeletConfiguration == nil || nodeClass.Spec.KubeletConfiguration.MaxPods == nil {
-		return template.Properties.NetworkInterfaces
-	}
-
 	// referring to: https://cloud.google.com/kubernetes-engine/docs/how-to/flexible-pod-cidr
-	maxPods := *nodeClass.Spec.KubeletConfiguration.MaxPods
+	maxPods := nodeClass.GetMaxPods()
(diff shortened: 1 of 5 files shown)


pkg/providers/instance/instance.go
```diff
@@ -440,12 +440,8 @@ func (p *DefaultProvider) buildInstance(nodeClaim *karpv1.NodeClaim, nodeClass *
 
 // nolint:gocyclo
 func (p *DefaultProvider) setupNetworkInterfaces(template *compute.InstanceTemplate, nodeClass *v1alpha1.GCENodeClass) []*compute.NetworkInterface {
-	if nodeClass.Spec.KubeletConfiguration == nil || nodeClass.Spec.KubeletConfiguration.MaxPods == nil {
```
jwcesign, line 443: I think we should keep this part

jwcesign (approved):

PR #141, opened 2025-10-18
fix: update list clusters apicall param

thameezb:
#### What type of PR is this?

/kind bug

Updates the params used in the listClusters call as Gcloud has seemingly dropped support for the deprecated arguments 

Encountered the auth failure on Friday evening, debugging since then, finally solved it live in my company cluster after this change

```release-note
NONE
```


jwcesign (approved):

jwcesign:
Thanks @thameezb 
Great Job!
/approve

PR #142, opened 2025-10-20
Set gke-provisioning label correctly in instance metadata

sthomson-wyn:
- set gke-provisioning label correctly in instance metadata

#### What type of PR is this?
/kind bug

This sets the gke-provisioning label to reflect the provisioning model

```release-note

```


jwcesign (commented):
Other lgtm

jwcesign:
Will merge after CI passes

PR #199, opened 2026-02-25
Fix instance leak race condition by resolving existing instances across all zones

igorkornienko:
#### What type of PR is this?

/kind bug

#### Description
This PR addresses a race condition in the instance creation logic that could result in leaked VMs in GCE.

#### The Problem:
When Karpenter attempts to provision an on-demand instance, it selects a random zone. If the `Instances.Insert` API call succeeds in GCE, but the response is lost (e.g., network timeout, controller crash, or a transient error), the controller fails to record the provider ID.

On the next reconciliation loop, the controller retries the creation. For on-demand instances, it might select a different random zone. Since it only checks for existence in the newly selected zone, it misses the instance already running in the previously selected zone. This results in the first instance being leaked (orphaned) while a second one is created.

#### The Solution:
I have updated the Create method in `instance.go` to perform a global check before attempting creation.
The controller now checks if an instance with the deterministic name already exists in any zone.
If an existing instance is found, it is immediately adopted and returned, preventing duplicate creation.

```Bug Fixes:
Fixed a race condition that could cause Karpenter to leak GCE instances.
```

### DIFF
--- a/pkg/providers/instance/instance.go
+++ b/pkg/providers/instance/instance.go
@@ -215,12 +215,15 @@ func (p *DefaultProvider) isInstanceExists(ctx context.Context, zone, instanceNa
 
 func (p *DefaultProvider) findInstanceByNodeClaim(ctx context.Context, nodeClaim *karpv1.NodeClaim) (*compute.Instance, error) {
 	instanceName := fmt.Sprintf("karpenter-%s", nodeClaim.Name)
-	call := p.computeService.Instances.AggregatedList(p.projectID)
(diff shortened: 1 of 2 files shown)


thameezb:
Thanks for the contribution, would it be possible to:
- add some unit-tests (where it makes sense)
- review this addition of code in light of https://github.com/cloudpilot-ai/karpenter-provider-gcp/pull/198, as the randomness is now dropped (it could still be created in a different zone, but we now respect constraints)

PR #215, opened 2026-03-14
feat: add subnet range name to GCENodeClass

### LINKED ISSUE #216: feat: allow overriding subnetwork secondary IP range for pod allocation in GCENodeClass
## Summary

Currently, the subnetwork secondary IPv4 range used for pod IP allocation is always inherited from the node pool instance template. This makes it impossible to use a custom pod CIDR range when provisioning nodes via Karpenter without modifying the underlying node pool template.

## Proposed Enhancement

Add a `subnetRangeName` field to `GCENodeClass` spec that allows users to explicitly specify the subnetwork secondary IPv4 range name for pod IP allocation.

**Example usage:**
```yaml
kind: GCENodeClass
metadata:
  name: default
spec:
  subnetRangeName: extended-pod
```

When set, this overrides the `SubnetworkRangeName` on the node's network interface alias IP range. When not set, behavior remains unchanged.

## Implementation

A functional implementation is available in PR #215.

cacarico:
Allow users to specify the subnetwork secondary IPv4 range for pod IP allocation via GCENodeClass. When set, overrides the range inherited from the node pool instance template.

#### What type of PR is this?

/kind feature
/kind api-change

Adds a new optional field `subnetRangeName` to `GCENodeClass` that allows users to override the subnetwork secondary IPv4 range used for pod IP allocation. Without this field, the range is inherited from the node pool instance template,
which may not be suitable for clusters with custom pod CIDR configurations.

Fixes #216 

The field maps directly to `SubnetworkRangeName` on the GCE `AliasIpRange` for the node's network interface. When not set, behavior is unchanged — the range inherited from the node pool instance template is preserved.

```release-note
Added `subnetRangeName` field to `GCENodeClass` spec, allowing users to specify
the subnetwork secondary IPv4 range for pod IP allocation, overriding the default
inherited from the node pool instance template.

### DIFF
--- a/charts/karpenter/crds/karpenter.k8s.gcp_gcenodeclasses.yaml
+++ b/charts/karpenter/crds/karpenter.k8s.gcp_gcenodeclasses.yaml
@@ -352,6 +352,15 @@ spec:
                   type: string
                 maxItems: 20
                 type: array
+              subnetRangeName:
+                description: |-
+                  SubnetRangeName is the name of the subnetwork secondary IPv4 range from which
+                  to allocate pod IP addresses. If not specified, the range inherited from the node
+                  pool instance template is used (typically the default pods range).
+                maxLength: 63
+                minLength: 1
+                pattern: ^[a-z]([-a-z0-9]{0,61}[a-z0-9])?$
+                type: string
               serviceAccount:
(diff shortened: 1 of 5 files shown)


thameezb (approved):
LGTM, lets resolve the merge conflicts and regenerate the files

PR #149, opened 2025-11-11
feat: add private nodes support via networkConfig

pedro-tamandua:
Add NetworkConfig to GCENodeClass spec to control external IP allocation and subnet selection for provisioned nodes.

Features:
- Add NetworkConfig struct with enableExternalIPAccess and subnetwork fields
- Implement resolveAccessConfigs() to conditionally disable external IPs
- Implement resolveSubnetwork() to allow custom subnet configuration
- Update CRD with new networkConfig fields
- Simplify CRD validation rules to meet Kubernetes cost limits
- Add unit tests for network configuration resolution
- Add example manifests for private nodes and custom subnets

When enableExternalIPAccess is false, nodes are provisioned without external IP addresses (private nodes), improving security posture.

## Usage Example
apiVersion: karpenter.k8s.gcp/v1alpha1
kind: GCENodeClass
metadata:
  name: private-nodes
spec:
  networkConfig:
    enableExternalIPAccess: false
  imageSelectorTerms:
    - alias: ContainerOptimizedOS@latest## Testing
- Unit tests added for `resolveAccessConfigs` and `resolveSubnetwork`
- All tests passing (5/5)
- Manually tested in GKE cluster - nodes provisioned without external IPs

## Backward Compatibility
Fully backward compatible. When `networkConfig` is not specified, behavior remains unchanged (nodes get external IPs by default).

PR #229, opened 2026-03-28
feat: add NetworkConfig for per-interface network settings (private nodes)

### LINKED ISSUE #182: Feature Request: Support disabling public IP addresses on provisioned nodes
### Description
Currently, the GCENodeClass specification does not provide a way to control whether provisioned GCE instances have external/public IP addresses assigned. Nodes provisioned by Karpenter are automatically assigned public IPs, which may not be desired for security and cost optimization reasons.

### Use Case
Many organizations prefer to run their Kubernetes nodes without public IP addresses for:
- Enhanced security posture by reducing attack surface
- Cost savings (avoiding egress charges associated with public IPs)
- Compliance requirements that mandate private-only networking
- Use of Cloud NAT or other private networking solutions for outbound connectivity

### Proposed Solution
Add a field to the GCENodeClassSpec to control network interface configuration, similar to how other cloud providers handle this:

```yaml
kind: GCENodeClass
metadata:
  name: default
spec:
  # ... existing fields ...
  
  # Proposed new field
  networkInterfaces:
    - omitExternalIP: true  # or similar boolean to disable public IPs
```

Alternatively, this could be controlled through metadata or a dedicated boolean field like `assignPublicIP: false`.

### Current Workaround
Currently, there is no workaround within Karpenter configuration. Organizations must either accept public IPs on Karpenter-provisioned nodes or use external automation to remove them after creation.

### Additional Context
- GCP supports creating instances without external IPs by omitting or providing an empty `accessConfigs` array in the network interface configuration
- Other Karpenter providers (AWS, Azure) provide similar controls for public IP assignment
- This feature would align with GCP best practices for private GKE clusters

### References
- GCP Documentation on creating instances without external IPs: https://cloud.google.com/compute/docs/instances/create-start-instance#no-external-ip
- Current Karpenter GCP implementation only exposes `NetworkTags` in the spec

dm3ch:
#### What type of PR is this?

/kind feature

Redesigns and completes the work started by @pedro-tamandua in #149 — adds `networkConfig` to `GCENodeClassSpec` so users can override per-network-interface settings on provisioned nodes without modifying the node pool template.

Pedro's original implementation used a flat `NetworkConfig` (a single `enableExternalIPAccess` + `subnetwork` at the spec level). This PR promotes the design to a **per-interface list** so multi-homed nodes with multiple NICs can be controlled independently, and fixes two correctness gaps found during review:

1. **AccessConfigs were not deep-copied** — the constructed NetworkInterface shared the underlying slice with the template, making future template mutations silently corrupt in-flight API requests.
2. **Missing `ForceSendFields`** — when `enableExternalIPAccess: false` cleared the AccessConfigs, the empty slice was not force-sent in the GCP API request body. On cluster configurations where the template's `ForceSendFields` did not already include `"AccessConfigs"`, the Compute API could default-insert `ONE_TO_ONE_NAT`, silently giving the private node a public IP.

**The primary use case is private nodes**: setting `enableExternalIPAccess: false` on the primary interface removes the `AccessConfig` entry, resulting in a node with no external (public) IP. Egress to the internet is still possible via Cloud NAT.

**Key changes:**

- New `NetworkConfig` struct with a `NetworkInterfaces []NetworkInterface` field added to `GCENodeClassSpec`
- `NetworkInterface` supports two optional overrides:
  - `enableExternalIPAccess` (`*bool`) — when `false`, clears the interface's `AccessConfigs` (disables public IP); when unset or `true`, inherits from the node pool template
  - `subnetwork` (string) — overrides the subnetwork URL for the interface; when unset, inherits from the node pool template
- Interfaces are matched to node pool template interfaces by position (index 0 = primary interface)
- Deep-copy of `AccessConfigs` on each interface (consistent with existing `AliasIpRanges` deep-copy)
- Force-send empty `AccessConfigs` when disabling external IP to prevent GCP API from defaulting to `ONE_TO_ONE_NAT`
- Existing `networkTags` and `subnetRangeName` fields are unchanged (backward compatible)
- CRD and deepcopy regenerated; 10 unit tests added; example YAML added

Closes #182
Replaces #149

**Not tested on a real GCP cluster** — I currently have no access to a GCP/GKE environment to do an end-to-end test. The logic is straightforward: when `enableExternalIPAccess: false` is set, the code sets `accessConfigs = nil` and forces it to be serialised as an empty array in the API request body (to prevent the Compute API from defaulting to `ONE_TO_ONE_NAT`). The unit tests cover all the relevant cases including the `ForceSendFields` behaviour.

Credit: the `NetworkConfig` concept and initial implementation originated in #149 by @pedro-tamandua. This PR redesigns the API to be per-interface and addresses the correctness gaps above.

```release-note
Add networkConfig field to GCENodeClass for per-interface network settings. Setting enableExternalIPAccess: false provisions private nodes without external IP addresses.
```


dm3ch:
## E2E test results

Tested on `karpenter-e2e-cluster` / `us-central1-f` with code from this branch (`d989d443`).

| Suite | Result | Specs |
|---|---|---|
| consolidation | ✅ PASS | 1/1 |
| drift | ✅ PASS | 2/2 |
| expiration | ✅ PASS | 1/1 |
| gc | ⏭ skipped | 0/1 (pending spec) |
| networking (private node, `enableExternalIPAccess: false`) | ✅ PASS | 1/1 |
| provisioning (amd64 on-demand, amd64 spot, arm64) | ✅ PASS | 5/5 |

All active specs pass. The networking test confirms that a node provisioned with `enableExternalIPAccess: false` starts successfully with no external IP, with outbound access via Cloud NAT.

docs/private-nodes.md
```diff
@@ -0,0 +1,109 @@
...
+This page explains how to control external IP assignment and subnetwork placement for nodes provisioned by Karpenter.
+
+## How Karpenter manages node pool templates
+
+When Karpenter starts, it creates two zero-node GKE node pools that act as instance templates:
+
+| Pool name | Image |
+|-----------|-------|
+| `karpenter-default` | Container-Optimized OS |
+| `karpenter-ubuntu` | Ubuntu |
```
thameezb: nit: Please add the two new ARM pools here and update the docs to reflect the 4 pools (have approved this for now, and you can add this in a doc PR)

thameezb (approved):

docs/private-nodes.md
dm3ch: Fixed in 2db0421a474ac7c889d5ea8e0c8166380ce86153

PR #236, opened 2026-04-02
fix(arm64): use arch-native node pool templates instead of patching kube-env

### LINKED ISSUE #235: ARM node provisioning broken: SERVER_BINARY_TAR_HASH not patched, use architecture-native templates instead
## Summary

ARM node provisioning (t2a, c4a instances) is broken despite PR #219. Nodes are created by GCP but **never register with the Kubernetes cluster**. The root cause is in the kube-env patching approach adopted by #219.

## Root cause

PR #219 patches `SERVER_BINARY_TAR_URL` in kube-env (replacing `linux-amd64` with `linux-arm64`) but does **not** patch `SERVER_BINARY_TAR_HASH`.

The instance template is based on an AMD64 node pool, so after patching, the ARM instance gets:

```
SERVER_BINARY_TAR_URL: .../kubernetes-server-linux-arm64.tar.gz   ✅ patched
SERVER_BINARY_TAR_HASH: f5e54065c9e...                            ❌ still the AMD64 sha512
```

The `configure.sh` GKE bootstrap script verifies the downloaded tarball against this hash:

```bash
if [[ -n "${SERVER_BINARY_TAR_HASH:-}" ]]; then
    local -r server_binary_tar_hash="${SERVER_BINARY_TAR_HASH}"   # uses wrong amd64 hash
else
    # downloads correct .sha512 from GKE release server
    download-or-bust "" "${server_binary_tar_urls[@]/.tar.gz/.tar.gz.sha512}"
fi
```

Since the hash is present (but wrong for the arm64 binary), checksum verification fails. The bootstrap script aborts, kubelet never starts, and the node never registers with the cluster.

## Why the patching approach is the wrong fix

The kube-env patching approach (`strings.ReplaceAll(kubeEnv, "linux-amd64", "linux-arm64")`) is fundamentally fragile:

1. `configure.sh` is a ~5000-line GKE-internal script. It contains many architecture-specific values: CNI, NPD, crictl, gcfsd, riptide snapshotter binaries — all with their own hashes. We can't reliably patch all of them.
2. `configure.sh` already handles architecture detection correctly via `uname -m` at runtime (for CNI, NPD, etc.). The only thing we break is the pre-set `SERVER_BINARY_TAR_HASH`.
3. GKE can add new architecture-specific fields in any release without warning.
4. String replacement on an internal config format is fragile by nature.

## Correct fix: architecture-aware template selection

GKE's node pool templates already contain fully correct, architecture-appropriate configuration when a node pool is created with an ARM machine type (t2a, c4a). The fix is:

1. Remove the `patchKubeEnvForArch` function (the `linux-amd64` → `linux-arm64` URL/hash replacement)
2. When provisioning an ARM instance, select an instance template from an ARM-capable node pool (one whose kube-env has `arch=arm64`) rather than the default AMD64 template
3. Return a clear error if no ARM template is found: *"ARM instance provisioning requires an ARM node pool (t2a or c4a) to exist in the cluster"*

This makes ARM provisioning work correctly and keeps validation in the hands of GKE (which owns the template format), not karpenter (which doesn't).

## Reproduction

Provision a `t2a-standard-2` instance using karpenter v0.2.x+ on any GKE cluster that only has AMD64 node pools. The node will be created by GCP (RUNNING) but will never appear in `kubectl get nodes` and will be terminated by karpenter after the registration timeout (~15 min).

Verified on GKE v1.35.1-gke.1396002.

Related to #214 #219

dm3ch:
#### What type of PR is this?

/kind bug
/kind regression

PR #219 (merged as ca3ec7f7) attempted to fix ARM64 node bootstrapping by patching `SERVER_BINARY_TAR_URL` in kube-env (`linux-amd64` → `linux-arm64`), but left `SERVER_BINARY_TAR_HASH` pointing to the AMD64 sha512. GKE's `configure.sh` downloads the binary and verifies it against this hash — the checksum mismatch causes ARM64 nodes to abort bootstrap silently, so they never join the cluster.

**Root cause:** `patchKubeEnvForArch` changed the download URL but not the corresponding hash, making ARM64 nodes permanently unbootstrappable.

**Fix:** Instead of patching kube-env, the controller now auto-creates dedicated ARM64 node pool templates (`karpenter-cos-arm64`, `karpenter-ubuntu-arm64`). GKE automatically populates the correct `SERVER_BINARY_TAR_URL`, `SERVER_BINARY_TAR_HASH`, and architecture labels in these templates. Instance creation selects the appropriate template based on the instance type's architecture label.

The `patchKubeEnvForArch` function introduced in #219 is removed entirely. `PatchKubeEnvForInstanceType` still patches `AUTOSCALER_ENV_VARS` labels (`arch=` and `machine-family=`) — just not the binary tarball URL/hash.

#### Design choices

**ARM64 template machine type — `c4a-standard-1`**

GKE needs an ARM64 machine type in the node pool config to generate ARM64-flavored kube-env (correct `SERVER_BINARY_TAR_URL`/hash, arch labels). The initial implementation used `t2a-standard-1`, but GCP has three ARM64 families with very different regional footprints:

| Family | Processor | Regions |
|--------|-----------|---------|
| `t2a` | Ampere Altra (Neoverse N1) | 3 (us-central1, europe-west4, asia-southeast1) |
| `n4a` | Google Axion (Neoverse N3) | ~9 |
| `c4a` | Google Axion (Neoverse V2) | ~28 |

Using `t2a-standard-1` as the sentinel would silently disable ARM64 provisioning in most GCP regions (us-east1, us-west1, europe-west1/2/3, asia-east1, …). `c4a-standard-1` is chosen because it has the broadest availability. This constant is used **only** for template generation — actual provisioned instances always use the machine type from the Karpenter instance type (`buildInstance` overrides it at line 669).

**Zone probe removed — `Locations` left nil**

The original implementation called `zonesForMachineType` to filter zones before creating the ARM64 node pool, to avoid a GCP error if the machine type was unavailable in some zones. This is unnecessary for two reasons:

1. The GKE API docs state explicitly: *"If Locations is unspecified during node pool creation, the Cluster.Locations value will be used."* Passing `nil` is valid and correct.
2. If `c4a` (or any ARM64 type) is genuinely unavailable in the region, the `nodePools.create` call returns an error, which is already handled gracefully — logged at Error level and ARM64 provisioning is skipped without affecting AMD64.

Fixes #235
Supersedes #219

- ARM64 node pools are created best-effort: if the region has no ARM64-capable zones the controller logs an error and skips ARM64 — AMD64 provisioning is unaffected.
- A single pair of ARM64 templates (`karpenter-cos-arm64`, `karpenter-ubuntu-arm64`) covers **all** ARM64 families (t2a, c4a, n4a) because `buildInstance` overrides the machine type at instance creation time from the Karpenter instance type.
- `utils_test.go` was introduced by #219; this PR updates assertions to match the new behaviour (URL/hash are no longer patched).

```release-note
fix(arm64): ARM64 nodes now bootstrap correctly. PR #219 introduced a kube-env URL patch that caused a checksum mismatch, silently preventing ARM64 nodes from joining the cluster. This PR replaces that approach with dedicated ARM64 node pool templates (karpenter-cos-arm64, karpenter-ubuntu-arm64) that GKE populates with the correct bootstrap configuration per architecture. All ARM64 machine families (t2a, c4a, n4a) are supported from a single template pair.
```


dm3ch:
## Review findings addressed

All findings from the hand-review have been fixed in commit [`33d4a900`](https://github.com/cloudpilot-ai/karpenter-provider-gcp/commit/33d4a900). Summary by category:

### Design fix: t2a → c4a, zone probe removed

The original implementation used `t2a-standard-1` as the ARM64 representative machine type and ran a per-zone availability probe at startup. Both were wrong:

**ARM64 family availability (GCP, April 2026):**
| Family | Processor | Regions |
|--------|-----------|---------|
| `t2a` | Ampere Altra (Neoverse N1) | 3 |
| `n4a` | Google Axion (Neoverse N3) | ~9 |
| `c4a` | Google Axion (Neoverse V2) | **~28** |

Using `t2a-standard-1` as the sentinel silently disabled ARM64 provisioning in the majority of GCP regions that only have `c4a`/`n4a` (us-east1, us-west1, europe-west1/2/3, asia-east1, …).

**Fix:** replaced with `c4a-standard-1` (unexported, clearly named `arm64MachineType`). The constant is used **only** to generate ARM64-flavored kube-env in the template — actual provisioned instance types are always overridden from the Karpenter instance type at `buildInstance:669`.

**Zone probe removed:** GKE API docs confirm that an unset `NodePool.Locations` defaults to the cluster's zones. If `c4a` is unavailable in a region, the `nodePools.create` call returns an error that's now logged at `Error` level and skipped gracefully.

A single ARM64 template pair (`karpenter-cos-arm64`, `karpenter-ubuntu-arm64`) covers **all** ARM64 families (t2a, c4a, n4a) because `buildInstance` overrides the machine type per-instance.

### Other fixes

- `resolveNodePoolName`: use `imagefamily.OSArchARM64Requirement` constant instead of bare `"arm64"` literal
- `resolveNodePoolName`: add `arch="" → "amd64"` default to match `kubeEnvPatchTargets` (guards against instance types with no explicit arch requirement)
- `tryCreateInstance`: extract `imageFamily` local variable (was called 3×)
- `GetInstanceTemplates` ARM64 loop: distinguish 404 (expected, pool absent → `V(1)`) from other errors (unexpected → `Error`)
- ARM64 pool creation failures promoted from `Info` to `Error`
- Removed stale "could be run concurrently" comment and two restatement-only inline comments
- Tightened `utils_test.go` comment wording

### e2e tests

`feat/e2e-tests` has been rebased onto the updated fix branch and ARM64 test selectors updated to include `c4a` ([force-pushed](https://github.com/dm3ch/karpenter-provider-gcp/tree/feat/e2e-tests)). e2e run in progress now.

thameezb (changes requested):

dm3ch:
Restored the four accidentally removed comments in `nodepooltemplate.go`. Thanks for catching that, @thameezb!

thameezb (approved):

PR #238, opened 2026-04-03
fix: pass capacityType into buildInstance to prevent spot/on-demand provisioning mismatch

### LINKED ISSUE #237: bug: buildInstance recomputes capacityType, provisioning wrong model and corrupting unavailable-offerings cache
**What happened**:

When a `NodeClaim` allows both spot and on-demand, `Create()` computes the capacity type once from *all* candidate instance types and passes it down the call chain. However, `buildInstance()` re-computed capacity type internally using only the *single* instance type being attempted. The two computations could disagree:

- `Create()` with instance types `[A(on-demand-only), B(spot+on-demand)]` → `spot`
- `buildInstance()` with just `A` → `on-demand`

**Primary consequence**: `A` is provisioned as on-demand when spot was requested — wrong pricing and eviction SLA, silently diverging from what the workload requires. The Karpenter node gets labelled `karpenter.sh/capacity-type=spot` (from `Create()`'s decision), but the underlying GCE instance runs as on-demand.

**Secondary consequence**: on an insufficient capacity error, the failure is cached under the `capacityType` from `Create()` (`A:zone:spot`), but the actual GCE call was on-demand. So `A:zone:on-demand` is never written to the cache, and future provisioning cycles keep retrying the same on-demand offering instead of skipping it until the TTL expires.

The provisioning model sent to the GCE API and the unavailable-offerings cache entry should both match the capacity type decided by `Create()`.

**How to reproduce it**:

1. Configure a `NodePool` with both `on-demand` and `spot` in `karpenter.sh/capacity-type` requirements.
2. Ensure the candidate instance type list contains at least one type with an available spot offering (so `Create()` resolves `capacityType=spot`) alongside another type that has no spot offering.
3. Trigger provisioning where the on-demand-only instance type is attempted in the `Create()` loop. Observe that the provisioned node carries the label `karpenter.sh/capacity-type=spot`, but the corresponding GCE instance has `scheduling.provisioningModel=STANDARD` (on-demand) — the node is billed at on-demand rates despite the spot label. If GCE returns an insufficient capacity error, it is additionally cached as `A:zone:spot` unavailable (the wrong key), leaving `A:zone:on-demand` unprotected so that on-demand for that type keeps being retried immediately on each subsequent cycle.

Fix: pass `capacityType` into `buildInstance()` as a parameter instead of re-computing it internally, so the GCE API call and the cache entry always agree.

**Environment**:

- Karpenter-provider-gcp version: all versions on `main`
- GKE version: any

dm3ch:
#### What type of PR is this?

/kind bug

`buildInstance` recomputed `capacityType` internally from the single instance type being attempted. `Create()` computes it from the full candidate list. These diverge when the chosen instance type has no spot offering but another candidate does — `Create()` returns `"spot"`, `buildInstance` returns `"on-demand"`.

Primary consequence: the wrong provisioning model is sent to the GCE API. A node requested as spot is silently provisioned as on-demand (higher cost, different eviction SLA). Secondary consequence: on insert failure, the wrong key is written to the unavailable-offerings cache, causing future scheduling cycles to mark the wrong offering exhausted.

**Fix:** remove the internal `getCapacityType` call in `buildInstance`; accept `capacityType` as a parameter so the GCE API call and the cache entry always agree with what `Create()` decided.

Fixes #237

Two-line change in `instance.go`:
1. Add `capacityType string` to `buildInstance`'s signature
2. Remove the internal `getCapacityType` recomputation

Test changes:
- Four `TestGetCapacityType_*` functions collapsed into one table-driven `TestGetCapacityType`
- `TestBuildInstance_UsesExternalCapacityTypeNotRecomputed`: calls `buildInstance` with `capacityType="spot"` and an on-demand-only instance type, asserts `ProvisioningModel==SPOT` — **fails on `origin/main`**, passes on this branch

```release-note
NONE
```

### DIFF
--- a/pkg/providers/instance/instance.go
+++ b/pkg/providers/instance/instance.go
@@ -383,7 +383,7 @@ func (p *DefaultProvider) getOrCreateInstance(ctx context.Context, nodeClaim *ka
 		return instance, false, nil
 	}
 
-	instance, err = p.buildInstance(nodeClaim, nodeClass, instanceType, template, nodePoolName, zone, instanceName)
+	instance, err = p.buildInstance(nodeClaim, nodeClass, instanceType, template, nodePoolName, zone, instanceName, capacityType)
 	if err != nil {
 		return nil, false, fmt.Errorf("building instance %s: %w", instanceName, err)
 	}
@@ -653,13 +653,11 @@ func (p *DefaultProvider) renderDiskProperties(instanceType *cloudprovider.Insta
 	return attachedDisks, nil
 }
 
-func (p *DefaultProvider) buildInstance(nodeClaim *karpv1.NodeClaim, nodeClass *v1alpha1.GCENodeClass, instanceType *cloudprovider.InstanceType, template *compute.InstanceTemplate, nodePoolName, zone, instanceName string) (*compute.Instance, error) {
+func (p *DefaultProvider) buildInstance(nodeClaim *karpv1.NodeClaim, nodeClass *v1alpha1.GCENodeClass, instanceType *cloudprovider.InstanceType, template *compute.InstanceTemplate, nodePoolName, zone, instanceName, capacityType string) (*compute.Instance, error) {
 	attachedDisks, err := p.renderDiskProperties(instanceType, nodeClass, zone)
 	if err != nil {
 		return nil, fmt.Errorf("rendering disk properties: %w", err)
 	}
-
-	capacityType := p.getCapacityType(nodeClaim, []*cloudprovider.InstanceType{instanceType})
 
 	// Setup metadata
 	if err := p.setupInstanceMetadata(template.Properties.Metadata, nodeClass, instanceType, nodeClaim, nodePoolName, capacityType); err != nil {
--- a/pkg/providers/instance/instance_test.go
+++ b/pkg/providers/instance/instance_test.go
@@ -798,3 +798,101 @@ func TestSetupScheduling_DoesNotMutateTemplate(t *testing.T) {
 	require.Equal(t, "MIGRATE", template.Properties.Scheduling.OnHostMaintenance,
 		"original template fields must be unchanged")
 }
+
+func spotOrOnDemandNodeClaim() *karpv1.NodeClaim {
+	return &karpv1.NodeClaim{
+		Spec: karpv1.NodeClaimSpec{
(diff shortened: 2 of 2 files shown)


thameezb (approved):

thameezb:
Conflicts

dm3ch:
@thameezb Fixed now

thameezb (changes requested):
The force pushed commit obviously is incorrect and wipes out other work entirely unrelated to the change.

My patience regarding these Claude commits are wearing thin, please review the code submitted yourself before tagging a reviewer, else any PRs from you will be blocked

dm3ch:
@thameezb Ok, let me check. I understand your frustration. At the same time, at the same time it's extremely hard to keep all PRs actualised when you have more than 10 ready-to-merge PRs without utilising LLM based tools :(

thameezb:
> @thameezb Ok, let me check. I understand your frustration. At the same time, at the same time it's extremely hard to keep all PRs actualised when you have more than 10 ready-to-merge PRs without utilising LLM based tools :(

Clearly they are not ready to be merged when there are conflicts and depedencies between them...a skill to learn as you go down this path is that quality is far greater than quantity.

There is no rush to fix every bug in this project, doing things in order of importance, and in steps will get us far further than just shot-gunning solutions all over the place 

To be clear, your contributions are welcome, but they need to be valuable

dm3ch:
Hi @thameezb — apologies for the messy history on this PR. The previous force-push was a mistake during rebasing that wiped unrelated commits; that's been fully resolved now. The branch is cleanly rebased onto main with two focused commits and presubmit is green.

On the question of whether this fix is valuable: the mismatch was caught during e2e test runs, where it caused an actual test failure. The root cause was a cache contamination effect: when the on-demand insert failed with insufficient capacity, the error was cached under the wrong key — `spot:X:zone` instead of `on-demand:X:zone`. As a result, `on-demand:X:zone` was never written to the unavailable-offerings cache, so the offering kept appearing valid in subsequent scheduling cycles. Each cycle would select it again, attempt provisioning, fail, and repeat — never converging. If I understand the code correctly, the provisioning model mismatch would also mean that any successfully-created node would be labeled as spot while actually running as on-demand — different cost and eviction SLA.

P.S. Also, I reached out to you via DM in Kubernetes slack org, to try to better understand your position

thameezb (approved):

PR #243, opened 2026-04-07
fix(instance): add idempotency guard and async delete for GCE instances

### LINKED ISSUE #242: Orphaned VM instances not deleted when Node/NodeClaim disappear
**What happened**:

We are observing GCE instances created by Karpenter that are not properly deleted and become either:

1. **Partially orphaned** (VM + Node still exist, but no NodeClaim)
2. **Fully orphaned** (VM exists in GCP, but no Node and no NodeClaim)

We investigated two concrete cases with both Karpenter logs and GCP audit logs.

---

### Case 1: Node still exists, NodeClaim deleted prematurely

Final observed state:
- VM exists in GCP
- Node still visible in Kubernetes
- Only DaemonSets running (eligible for consolidation)
- **NodeClaim is missing**

Karpenter logs:

```text
"message": "Deleting instance"
"message": "Delete operation started"
(repeated multiple times)
...
"message": "Instance already deleted or not found"
"message": "deleted nodeclaim"
```

GCP audit logs confirm:
- Instance was successfully created
- Karpenter issued `compute.instances.delete`
- Delete operation started (`status: RUNNING`)
- No confirmed successful deletion completion
- Later `notFound` responses observed

Observed behavior:
- Multiple overlapping delete operations issued for the same instance
- NodeClaim deleted even though instance was still alive
- Node remained in Kubernetes, but Karpenter lost ownership

---

### Case 2: Fully orphaned VM

Final observed state:
- VM exists in GCP
- Not visible in Kubernetes
- No NodeClaim
- No ownership by Karpenter

Karpenter logs:

```text
"message": "Delete operation started"
...
"error": "googleapi: Error 403: Too many pending operations on a resource., rateLimitExceeded"
...
"message": "Instance already deleted or not found"
"message": "deleted node"
"message": "deleted nodeclaim"
```

GCP audit logs confirm:
- Multiple `compute.instances.delete` requests issued for same instance
- Many delete operations in `PENDING` state simultaneously
- GCP returns rate limit errors
- Later `notFound` responses

Observed behavior:
- Delete requests are issued repeatedly without waiting for prior operations
- GCP rejects requests due to too many pending operations
- Controller still concludes instance is deleted
- Node + NodeClaim removed
- VM remains running → becomes permanent orphan

---

For node termination:

1. Node is drained
2. A single instance delete operation is initiated
3. Controller waits for the GCE operation to complete successfully
4. Only then are Node and NodeClaim finalized

Additionally:
- Delete operations should not overlap for the same instance
- Instances must not outlive their NodeClaim lifecycle

---

This issue appears intermittently but follows this pattern:

1. Karpenter provisions nodes normally
2. Node becomes eligible for deletion (e.g. consolidation or termination)
3. Provider initiates deletion
4. Multiple delete requests are issued concurrently
5. Either:
   - NodeClaim is deleted prematurely → Node still exists
   - Node + NodeClaim deleted → VM orphaned in GCP

---

From GCP audit logs:

- Delete operations are **asynchronous** and remain in `RUNNING` / `PENDING`
- Provider issues **multiple delete requests for the same instance**
- This leads to:
  - `403 Too many pending operations on a resource`
- Provider later treats:
  - `Instance already deleted or not found`
  as a terminal success condition

However, in both cases:
- The VM still existed afterward

This suggests:

The provider is not correctly tracking or waiting for async delete operations  
Finalization of Node / NodeClaim happens before deletion is actually confirmed  
Overlapping delete requests are causing inconsistent state and eventual orphaning

---

**Environment**:

- Karpenter-provider-gcp version (use `git describe --tags --dirty --always`): `v0.2.1`
- GKE version: v1.35.2-gke.1269001
- Others:
  - Region: EU and US clusters
  - Zones: multiple zones within above regions
  - Consolidation enabled (WhenEmpty and ConsolidateAfter)

dm3ch:
#### What type of PR is this?

/kind bug
/kind regression

`Delete()` previously called the GCE Instances.Delete API without first checking whether the instance was already being deleted. On repeated reconciles this sent multiple concurrent delete operations to the same instance, causing GCE to return `403 rateLimitExceeded`. The eventual spurious `404` was misinterpreted as "instance gone", causing Karpenter to remove the NodeClaim finalizer while the VM was still running — leaving orphaned instances.

Changes:
- **Idempotency guard**: GET the instance before issuing DELETE. If already `STOPPING`, return nil (delete in flight, re-queue naturally). If `TERMINATED` or 404, return `NodeClaimNotFoundError` so Karpenter finalizes the NodeClaim.
- **Async delete**: removed the blocking `waitOperationDone` call from `Delete()`. Under mass-delete scenarios this was saturating reconciler goroutines for up to 2 minutes per instance. The status-check loop handles completion detection on re-queue.
- **Tests**: four httptest-based unit tests covering the 404, STOPPING, TERMINATED, and RUNNING paths.

Fixes #242

The GET-before-DELETE approach is intentional over retry logic: it avoids sending duplicate operations at all rather than handling their errors. STOPPING is returned as nil (not an error) because delete is already in flight and the reconciler will re-check on the next cycle.

```release-note
Fixed duplicate GCE delete operations on repeated reconciles that caused 403 rateLimitExceeded errors, which could result in orphaned VM instances.
```

### DIFF
--- a/pkg/providers/instance/instance.go
+++ b/pkg/providers/instance/instance.go
@@ -968,17 +968,38 @@ func (p *DefaultProvider) getZonesInRegion(ctx context.Context) ([]string, error
 }
 
 func (p *DefaultProvider) Delete(ctx context.Context, providerID string) error {
-	project, zone, instance, err := parseGCEProviderID(providerID)
+	project, zone, instanceName, err := parseGCEProviderID(providerID)
 	if err != nil {
 		return fmt.Errorf("parsing provider ID: %w", err)
 	}
 
-	log.FromContext(ctx).Info("Deleting instance", "project", project, "zone", zone, "instance", instance)
+	log.FromContext(ctx).Info("Deleting instance", "project", project, "zone", zone, "instance", instanceName)
 
-	op, err := p.computeService.Instances.Delete(project, zone, instance).Context(ctx).Do()
+	// Check current state before issuing a delete. This prevents sending multiple overlapping
+	// delete operations to GCE (which cause 403 rateLimitExceeded on repeated reconciles).
+	inst, err := p.computeService.Instances.Get(project, zone, instanceName).Context(ctx).Do()
 	if err != nil {
 		if isInstanceNotFoundError(err) {
-			log.FromContext(ctx).Info("Instance already deleted or not found", "instance", instance)
+			log.FromContext(ctx).Info("Instance already deleted", "instance", instanceName)
+			return cloudprovider.NewNodeClaimNotFoundError(fmt.Errorf("instance not found: %w", err))
+		}
+		return fmt.Errorf("getting instance state before delete: %w", err)
+	}
+
+	switch inst.Status {
+	case InstanceStatusTerminated:
+		log.FromContext(ctx).Info("Instance already terminated", "instance", instanceName)
+		return cloudprovider.NewNodeClaimNotFoundError(fmt.Errorf("instance %s is already %s", instanceName, inst.Status))
+	case InstanceStatusStopping:
+		// Delete already in flight; caller re-queues and rechecks.
+		log.FromContext(ctx).Info("Instance delete already in progress", "instance", instanceName)
+		return nil
+	}
+
+	op, err := p.computeService.Instances.Delete(project, zone, instanceName).Context(ctx).Do()
+	if err != nil {
+		if isInstanceNotFoundError(err) {
+			log.FromContext(ctx).Info("Instance disappeared before delete call", "instance", instanceName)
 			return cloudprovider.NewNodeClaimNotFoundError(fmt.Errorf("instance not found: %w", err))
 		}
 		return fmt.Errorf("deleting instance: %w", err)
--- a/pkg/providers/instance/instance_test.go
+++ b/pkg/providers/instance/instance_test.go
@@ -18,17 +18,23 @@ package instance
 
 import (
 	"context"
+	"encoding/json"
+	"net/http"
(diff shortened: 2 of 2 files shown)


dm3ch:
## E2E Test Results ✅

All test suites passed against this PR (karpenter deployed from `fix/delete-sync-and-idempotency`, tests from `feat/e2e-tests`).

**Cluster:** `karpenter-e2e-cluster` · `us-central1-f` · GKE 1.35.1-gke.1396002

| Suite | Result | Duration |
|-------|--------|----------|
| Provisioning (amd64 on-demand) | ✅ PASSED | ~10 min |
| Provisioning (amd64 spot) | ✅ PASSED | ~10 min |
| Consolidation (amd64 on-demand) | ✅ PASSED | 3m47s |
| Drift (amd64 on-demand) | ✅ PASSED | ~12 min |
| Expiration (amd64 on-demand) | ✅ PASSED | ~12 min |
| NodeClaim GC (orphaned VM) | ✅ PASSED | ~8 min |

ARM64 tests are disabled (`XIt`) pending ARM64 fix in `main`.

**Notes:**
- Instance types restricted to `n2-standard-2`/`n2-standard-4` (e2 family has capacity issues in us-central1).
- Consolidation required a per-nodepool `NoSchedule` taint fix (committed to `feat/e2e-tests`) to prevent GKE system Deployments (`konnectivity-agent`, `kube-dns`) from landing on test nodes and resetting the consolidation timer.

thameezb (approved):

PR #244, opened 2026-04-07
feat(gc): add GC controller to delete orphaned GCE instances

### LINKED ISSUE #242: Orphaned VM instances not deleted when Node/NodeClaim disappear

dm3ch:
#### What type of PR is this?

/kind feature

Adds a singleton garbage-collection controller (`instance.garbagecollection`) that periodically reconciles GCE instances against known Kubernetes NodeClaims and deletes any that have no corresponding NodeClaim.

This closes the gap where an instance can outlive its NodeClaim due to: a missed delete call, a controller restart during teardown, or a NodeClaim deleted directly from the API server without going through Karpenter's termination path.

The controller:
- Lists all GCE instances via `CloudProvider.List()`
- Lists all `NodeClaim` objects in Kubernetes
- Deletes any instance older than 30 s that has no matching NodeClaim (by `ProviderID`)
- Skips instances already being deleted (`DeletionTimestamp` set)
- Logs and continues on per-instance delete errors (does not abort the full reconcile)
- Re-queues every 2 minutes

**Cluster-location label**

A new GCE label, `goog-k8s-cluster-location`, is stamped on every instance at creation time. This is checked alongside the existing `goog-k8s-cluster-name` label to distinguish clusters that share the same name in different GCP locations.

Both cluster-identity labels are stamped after the instance-type requirements loop so that user-supplied NodePool labels with the same key cannot overwrite them.

**Upgrade safety / backward compatibility**

Instances without `goog-k8s-cluster-location` (created by an older Karpenter version) are treated as belonging to this cluster in the instance cache — same behavior as before this PR. This means no nodes are lost on upgrade. However, the GC controller explicitly skips such instances, since without the label we cannot confirm they belong exclusively to this cluster. They remain under full Karpenter management until naturally replaced, at which point the new node picks up the label and becomes eligible for GC.

Once all clusters have migrated (all pre-existing nodes rotated), the backward-compatibility fallback can be dropped in a follow-up: tighten `belongsToCluster` to strict matching and add `goog-k8s-cluster-location` as a server-side GCE API filter. TODOs are left in the code at the relevant locations.

Fixes #242

This PR is independent of cloudpilot-ai/karpenter-provider-gcp#243 — it branches from `main` and can be reviewed/merged in any order. The 30-second grace period prevents newly created instances from being GC'd before their NodeClaim has been persisted to the API server.

The controller is named `instance.garbagecollection` (not `nodeclaim.garbagecollection`) to avoid a name collision with the same-named controller already registered by the upstream karpenter core.

**Migration:** no action required to upgrade. See [MIGRATION.md](MIGRATION.md) for optional steps to rotate existing nodes early and clean up any pre-existing orphaned instances using the provided `gcloud` commands.

```release-note
Added a garbage-collection controller that periodically detects and deletes GCE VM instances with no corresponding NodeClaim, preventing orphaned instances from accumulating.
Added goog-k8s-cluster-location GCE label to instances at creation time. The GC controller only operates on labeled instances, so nodes created by older Karpenter versions are never deleted on upgrade. No migration steps required. See MIGRATION.md for optional cleanup of pre-existing orphaned instances.
```


thameezb:
conflicts

dm3ch:
Rebased onto current main — merge conflict resolved. Ready for re-review.

thameezb:
Unfortunately another conflict

dm3ch:
> Unfortunately another conflict

Fixed that, let's hope it's the last time for this PR :)

thameezb (approved):

PR #245, opened 2026-04-07
fix(instance): skip zones with cached insufficient capacity in zone selection

dm3ch:
#### What type of PR is this?

/kind bug

`selectZone` did not consult the unavailability cache before picking a zone, causing two problems:

1. **On-demand**: zones with cached insufficient capacity entries were retried on every provisioning attempt, calling `MarkUnavailable` each time and resetting the 30-minute TTL. Zones would stay poisoned indefinitely instead of recovering after 30 minutes.

2. **Spot**: `cheapestCompatibleZone` filtered by `offering.Available` (an indirect read of the same cache), so spot happened to be protected — but only as a side-effect of the price-selection logic, not explicitly.

**Fix**: add a single `IsUnavailable` pre-filter before the `capacityType` branch so both on-demand and spot skip zones with cached insufficient capacity entries via the same code path. Once a zone is skipped (not retried), its TTL ticks down naturally and the zone recovers after 30 minutes. If all zones for an instance type are exhausted, `selectZone` returns `InsufficientCapacityError` immediately instead of making a guaranteed-to-fail GCP API call.

The now-redundant `offering.Available` check has been removed from `cheapestCompatibleZone` since zones are pre-filtered before the call.

**Relationship to #238**: PR #238 fixes `capacityType` being mis-computed in `buildInstance`, which caused on-demand insufficient capacity errors to be cached under the wrong key (`spot` instead of `on-demand`). This PR is independent and correct with or without #238 merged, but the two together fully fix the unavailability cache correctness.

Related to #237

N/A

```release-note
Fixed on-demand zone selection to skip zones with cached insufficient capacity entries, preventing indefinite TTL refresh and allowing zones to recover naturally after 30 minutes.
```

### DIFF
--- a/pkg/providers/instance/instance.go
+++ b/pkg/providers/instance/instance.go
@@ -522,11 +522,9 @@ func (p *DefaultProvider) selectZone(ctx context.Context, nodeClaim *karpv1.Node
 	// Without this filter, every retry calls MarkUnavailable and resets the
 	// 30-min TTL, preventing natural expiry. Applied to both on-demand and
 	// spot so both paths behave consistently.
-	if p.unavailableOfferings != nil {
-		zones = lo.Filter(zones, func(z string, _ int) bool {
-			return !p.unavailableOfferings.IsUnavailable(instanceType.Name, z, capacityType)
-		})
-	}
+	zones = lo.Filter(zones, func(z string, _ int) bool {
+		return !p.unavailableOfferings.IsUnavailable(instanceType.Name, z, capacityType)
+	})
 	if len(zones) == 0 {
 		return "", cloudprovider.NewInsufficientCapacityError(
 			fmt.Errorf("all zones exhausted for instance type %s", instanceType.Name))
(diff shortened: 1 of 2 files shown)


pkg/providers/instance/instance.go
```diff
@@ -515,12 +512,21 @@ func (p *DefaultProvider) selectZone(ctx context.Context, nodeClaim *karpv1.Node
 		return "", fmt.Errorf("no zones match topology requirement %q", corev1.LabelTopologyZone)
 	}
 
-	// For on-demand, randomly select a zone from those that satisfy the requirement,
```
thameezb, line 521: nit: lets not remove comments that are not impacted by this change

pkg/providers/instance/instance.go
```diff
@@ -515,12 +512,21 @@ func (p *DefaultProvider) selectZone(ctx context.Context, nodeClaim *karpv1.Node
...
+	})
+	if len(zones) == 0 {
+		return "", cloudprovider.NewInsufficientCapacityError(
+			fmt.Errorf("all zones exhausted for instance type %s", instanceType.Name))
+	}
+
 	if capacityType == karpv1.CapacityTypeOnDemand {
 		return randomZone(zones), nil
 	}
-	// else for spot, choose the cheapest zone
```
thameezb, line 526: nit: lets not remove comments that are not impacted by this change

thameezb (changes requested):

dm3ch:
Restored the accidentally removed comments in `selectZone`. Thanks for catching that, @thameezb!

thameezb (approved):

PR #251, opened 2026-04-08
feat: add shieldedInstanceConfig support to GCENodeClass

sleterrier:
#### What type of PR is this?

/kind feature
/kind api-change

#### What this PR does / why we need it:

Adds `shieldedInstanceConfig` field to the `GCENodeClass` spec, enabling Shielded VM configuration (Secure Boot, vTPM, integrity monitoring) on Karpenter-provisioned instances.

GCP organizations enforcing the `constraints/compute.requireShieldedVm` organization policy cannot use karpenter-gcp-provider today because instances are created without `ShieldedInstanceConfig`, resulting in a `412 conditionNotMet` error:
```
googleapi: Error 412: Constraint constraints/compute.requireShieldedVm violated for project ...
Secure Boot is not enabled in the 'shielded_instance_config' field.
```

Usage:
```yaml
apiVersion: karpenter.k8s.gcp/v1alpha1
kind: GCENodeClass
metadata:
  name: shieldedInstance-example
spec:
  shieldedInstanceConfig:
    enableSecureBoot: true
    enableVtpm: true
    enableIntegrityMonitoring: true
```

Changes:
- Add `ShieldedInstanceConfig` struct and optional field to `GCENodeClassSpec`
- Add deep copy functions for `ShieldedInstanceConfig`
- Set `ShieldedInstanceConfig` on `compute.Instance` in `buildInstance()`, using `ForceSendFields` to ensure boolean zero values are sent
- Add `shieldedInstanceConfig` to CRD schema

#### Does this PR introduce a user-facing change?

```release-note
Add shieldedInstanceConfig field to GCENodeClass spec, allowing configuration of Shielded VM options (Secure Boot, vTPM, integrity monitoring).
```

PR #253, opened 2026-04-09
feat: add CMEK and Secure Boot settings to template node pools

sleterrier:
#### What type of PR is this?

/kind feature

The `nodepooltemplate` controller auto-creates `karpenter-default` and `karpenter-ubuntu` GKE node pools, but uses a minimal `NodeConfig` that lacks boot disk encryption and Shielded VM settings. This causes auto-creation to fail in environments that enforce `gcp.restrictNonCmekServices` or `compute.requireShieldedVm` org policy constraints.

This PR adds two new controller settings:

- **`DEFAULT_NODEPOOL_BOOT_DISK_KMS_KEY`** — Cloud KMS key for boot disk encryption on template node pools
- **`DEFAULT_NODEPOOL_SHIELDED_VM`** — enables Secure Boot on template node pools

Both settings are optional, backward-compatible, and follow the existing `DEFAULT_NODEPOOL_SERVICE_ACCOUNT` pattern.

Helm usage:

```yaml
controller:
  settings:
    defaultNodePoolBootDiskKmsKey: "projects/my-project/locations/us-central1/keyRings/my-ring/cryptoKeys/my-key"
    defaultNodePoolShieldedVM: true
```

When `defaultNodePoolShieldedVM` is `false` (default), we still send a `ShieldedInstanceConfig` struct with `EnableSecureBoot: false`. Since the field uses `omitempty`, it gets omitted from the JSON request and the GKE API uses its default. Same applies to `BootDiskKmsKey` with an empty string. 
I opted for this approach to keep the code minimal, but happy to add explicit nil-checking if you'd prefer.

```release-note
Add DEFAULT_NODEPOOL_BOOT_DISK_KMS_KEY and DEFAULT_NODEPOOL_SHIELDED_VM controller settings to configure boot disk encryption and Secure Boot on auto-created template node pools.
```

### DIFF
--- a/pkg/providers/nodepooltemplate/nodepooltemplate.go
+++ b/pkg/providers/nodepooltemplate/nodepooltemplate.go
@@ -40,9 +40,11 @@ type Provider interface {
 }
 
 type DefaultProvider struct {
+	bootDiskKmsKey        string
 	computeService        *compute.Service
 	containerService      *container.Service
 	kubeClient            client.Client
+	shieldedVM            bool
 	versionProvider       version.Provider
 	ClusterInfo           ClusterInfo
 	defaultServiceAccount string
@@ -76,7 +78,8 @@ const arm64MachineType = "c4a-standard-1"
 
 func NewDefaultProvider(ctx context.Context, kubeClient client.Client, computeService *compute.Service,
 	containerService *container.Service, versionProvider version.Provider,
-	clusterName, region, projectID, serviceAccount, clusterLocation, nodeLocation string) *DefaultProvider {
+	clusterName, region, projectID, serviceAccount, clusterLocation, nodeLocation, bootDiskKmsKey string,
+	shieldedVM bool) *DefaultProvider {
 
 	zones, err := resolveZones(ctx, computeService, projectID, region)
 	if err != nil {
@@ -85,9 +88,11 @@ func NewDefaultProvider(ctx context.Context, kubeClient client.Client, computeSe
 	}
 
 	return &DefaultProvider{
+		bootDiskKmsKey:        bootDiskKmsKey,
 		kubeClient:            kubeClient,
 		computeService:        computeService,
 		containerService:      containerService,
+		shieldedVM:            shieldedVM,
 		versionProvider:       versionProvider,
 		defaultServiceAccount: serviceAccount,
 		ClusterInfo: ClusterInfo{
@@ -181,11 +186,17 @@ func (p *DefaultProvider) ensureKarpenterNodePoolTemplate(ctx context.Context, i
 	}
 
 	nodeConfig := &container.NodeConfig{
+		BootDiskKmsKey: p.bootDiskKmsKey,
 		ImageType:      imageType,
 		ServiceAccount: serviceAccount,
 	}
 	if machineType != "" {
 		nodeConfig.MachineType = machineType
+	}
+	if p.shieldedVM {
+		nodeConfig.ShieldedInstanceConfig = &container.ShieldedInstanceConfig{
+			EnableSecureBoot: true,
+		}
 	}
 	nodePoolOpts := &container.CreateNodePoolRequest{
 		NodePool: &container.NodePool{
--- a/charts/karpenter/templates/deployment.yaml
+++ b/charts/karpenter/templates/deployment.yaml
@@ -53,6 +53,14 @@ spec:
               value: {{ .Values.controller.settings.projectID }}
             - name: CLUSTER_LOCATION
               value: {{ .Values.controller.settings.clusterLocation }}
+            {{- if .Values.controller.settings.defaultNodePoolBootDiskKmsKey }}
+            - name: DEFAULT_NODEPOOL_BOOT_DISK_KMS_KEY
+              value: {{ .Values.controller.settings.defaultNodePoolBootDiskKmsKey }}
+            {{- end }}
+            {{- if .Values.controller.settings.defaultNodePoolShieldedVM }}
+            - name: DEFAULT_NODEPOOL_SHIELDED_VM
+              value: "true"
+            {{- end }}
             {{- if .Values.controller.settings.nodeLocation }}
             - name: NODE_LOCATION
               value: {{ .Values.controller.settings.nodeLocation }}
--- a/charts/karpenter/values.yaml
+++ b/charts/karpenter/values.yaml
@@ -74,11 +74,17 @@ controller:
     projectID: ""
     # -- The GCP region for instance type discovery (e.g., us-central1)
     clusterLocation: ""
+    # -- The GCP cluster name.
+    clusterName: ""
+    # -- Cloud KMS key for boot disk encryption on default node pool templates.
+    # Format: projects/{project}/locations/{location}/keyRings/{keyring}/cryptoKeys/{key}
+    defaultNodePoolBootDiskKmsKey: ""
+    # -- Enable Shielded VM (Secure Boot) on default node pool templates.
+    # Required when the compute.requireShieldedVm org policy is enforced.
+    defaultNodePoolShieldedVM: false
     # -- The exact GCP cluster location for GKE API calls (e.g., us-central1-a for zonal, us-central1 for regional).
     # If not set, defaults to 'clusterLocation' for backward compatibility.
     nodeLocation: ""
-    # -- The GCP cluster name.
-    clusterName: ""
     # -- The VM memory overhead as a percent that will be subtracted from the total memory for all instance types. The value of `0.075` equals to 7.5%.
     vmMemoryOverheadPercent: 0.065
     # -- The maximum length of a batch window. The longer this is, the more pods we can consider for provisioning at one
--- a/pkg/operator/operator.go
+++ b/pkg/operator/operator.go
@@ -103,6 +103,8 @@ func NewOperator(ctx context.Context, operator *operator.Operator) (context.Cont
 		options.FromContext(ctx).NodePoolServiceAccount,
 		options.FromContext(ctx).ClusterLocation,
 		options.FromContext(ctx).NodeLocation,
+		options.FromContext(ctx).NodePoolBootDiskKmsKey,
+		options.FromContext(ctx).NodePoolShieldedVM,
 	)
 	imageProvider := imagefamily.NewDefaultProvider(computeService, nodeTemplateProvider)
 	pricingProvider, err := pricing.NewDefaultProvider(ctx, region)
--- a/pkg/operator/options/options.go
+++ b/pkg/operator/options/options.go
@@ -40,8 +40,12 @@ const (
 	vmMemoryOverheadPercentFlagName   = "vm-memory-overhead-percent"
 	gkeEnableInterruption             = "INTERRUPTION"
 	GCPAuth                           = "GOOGLE_APPLICATION_CREDENTIALS"
+	nodePoolBootDiskKmsKeyEnvVarName  = "DEFAULT_NODEPOOL_BOOT_DISK_KMS_KEY"
+	nodePoolBootDiskKmsKeyFlagName    = "default-nodepool-boot-disk-kms-key"
 	nodePoolServiceAccountEnvVarName  = "DEFAULT_NODEPOOL_SERVICE_ACCOUNT"
 	nodePoolServiceAccountFlagName    = "default-nodepool-service-account"
+	nodePoolShieldedVMEnvVarName      = "DEFAULT_NODEPOOL_SHIELDED_VM"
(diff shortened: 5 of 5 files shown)


pkg/providers/nodepooltemplate/nodepooltemplate.go
```diff
@@ -175,8 +180,12 @@ func (p *DefaultProvider) ensureKarpenterNodePoolTemplate(ctx context.Context, i
 			InitialNodeCount: 0,
 			Locations:        p.ClusterInfo.Zones,
 			Config: &container.NodeConfig{
+				BootDiskKmsKey: p.bootDiskKmsKey,
 				ImageType:      imageType,
 				ServiceAccount: serviceAccount,
+				ShieldedInstanceConfig: &container.ShieldedInstanceConfig{
+					EnableSecureBoot: p.shieldedVM,
+				},
```
dm3ch: The `ShieldedInstanceConfig` struct is always allocated (even when `shieldedVM` is `false`), which means the GKE API always receives `"shieldedInstanceConfig": {}` in the request body — rather than the field being absent.

The PR description says `omitempty` will drop it, but `omitempty` applies to the _fields inside_ the struct, not to the struct pointer itself. A non-nil pointer is never omitted, so the parent key is always serialised.

The risk is that sending an explicit-but-empty `ShieldedInstanceConfig` may cause GKE to treat unset inner fields as `false` — which would **disable** vTPM and Integrity Monitoring even though both are `true` by default in GCP (and in Terraform's provider):

> [`enable_vtpm`](https://registry.terraform.io/providers/hashicorp/google/latest/docs/resources/container_node_pool#enable_vtpm) — Defaults to `true`.
> [`enable_integrity_monitoring`](https://registry.terraform.io/providers/hashicorp/google/latest/docs/resources/container_node_pool#enable_integrity_monitoring) — Defaults to `true`.
> [`enable_secure_boot`](https://registry.terraform.io/providers/hashicorp/google/latest/docs/resources/container_node_pool#enable_secure_boot) — Defaults to `false`.

A nil-guard makes the behaviour unambiguous — only set the struct when the user explicitly opts in:

```go
if p.shieldedVM {
    cfg.ShieldedInstanceConfig = &container.ShieldedInstanceConfig{
        EnableSecureBoot: true,
    }
}
```

This way, when `shieldedVM` is `false` the field is fully absent from the request and GKE applies all its defaults (vTPM on, Integrity Monitoring on, Secure Boot off) — matching what both the GCP Console and Terraform produce out of the box.

dm3ch (commented):
Thanks for the PR! The approach is solid and the default values (Secure Boot off, no CMEK) correctly match GCP Console / Terraform defaults. One concern about how the `ShieldedInstanceConfig` struct is constructed — see the inline comment below.

pkg/providers/nodepooltemplate/nodepooltemplate.go
```diff
@@ -175,8 +180,12 @@ func (p *DefaultProvider) ensureKarpenterNodePoolTemplate(ctx context.Context, i
...
 			},
 		},
 	}
```
dm3ch: `ShieldedInstanceConfig` is always allocated, so the parent key is always present in the API request even when `shieldedVM` is `false`. A non-nil pointer is never dropped by `omitempty`, so the GKE API receives `"shieldedInstanceConfig": {}`, which may override the server-side defaults for vTPM and Integrity Monitoring (both default to `true` in [GCP Console and Terraform](https://registry.terraform.io/providers/hashicorp/google/latest/docs/resources/container_node_pool#nested_shielded_instance_config)).

Setting it only when the user opts in makes the intent unambiguous:

```suggestion
			Config: &container.NodeConfig{
				BootDiskKmsKey: p.bootDiskKmsKey,
				ImageType:      imageType,
				ServiceAccount: serviceAccount,
			},
		},
	}
	if p.shieldedVM {
		nodePoolOpts.NodePool.Config.ShieldedInstanceConfig = &container.ShieldedInstanceConfig{
			EnableSecureBoot: true,
		}
	}
```

dm3ch (commented):
Concrete suggestion for the `ShieldedInstanceConfig` nil-guard (complements the previous review comment).

dm3ch:
From my understanding, this change may not align with the long-term vision for the project — it feels like these settings would fit better as `GCENodeClass` fields rather than global controller options, but happy to be corrected if there's reasoning I'm missing.

There also seems to be an implementation concern: when `shieldedVM` is `false`, the code still allocates `&container.ShieldedInstanceConfig{EnableSecureBoot: false}` and assigns it. A non-nil pointer is never dropped by `omitempty`, so the GKE API would always receive `"shieldedInstanceConfig": {}` — which might cause it to treat `enableVtpm` and `enableIntegrityMonitoring` as explicitly `false`, potentially disabling features that GCP enables by default. See the inline comment for a suggested fix.

**The split concern**

As I understand it, `GCENodeClass` already carries `spec.disks[].kmsKeyName` for CMEK on Karpenter-provisioned VMs. This PR adds a parallel `DEFAULT_NODEPOOL_BOOT_DISK_KMS_KEY` controller option for the template node pools, which means a user enforcing CMEK would need to express the same intent in two different places — Helm values for the template pools, and GCENodeClass for actual nodes.

Shielded VM seems to have the same split in the opposite direction: the PR covers template pools but GCENodeClass has no `shieldedInstanceConfig` field yet, so actual Karpenter-provisioned VMs would remain uncovered.

**How karpenter-aws approaches this**

From what I can see in karpenter-provider-aws, all security-related configuration (KMS key IDs, IMDSv2 options, encryption flags) lives in `EC2NodeClass`. Controller options appear to be strictly cluster identity and feature flags, with no security settings at the operator level. AWS Karpenter synthesises EC2 launch templates per-NodeClass, giving a single source of truth.

**A possible direction**

It seems like a more consistent shape for GCP might be:

1. Add `spec.shieldedInstanceConfig` (with `enableSecureBoot`, `enableVtpm`, `enableIntegrityMonitoring`) to `GCENodeClass`
2. Drive the template node pool configuration from the GCENodeClasses that reference it, rather than a global controller flag
3. Drop `DEFAULT_NODEPOOL_BOOT_DISK_KMS_KEY` / `DEFAULT_NODEPOOL_SHIELDED_VM` from operator options

That way CMEK and Shielded VM would be configured once in GCENodeClass and apply to both the template pool and the actual VMs — and different node classes could use different KMS keys if needed. But I appreciate this is a larger change than the PR intends.

pkg/providers/nodepooltemplate/nodepooltemplate.go
sleterrier: Good catch! I initially had a nil-guard but refactored it away to keep the code more compact, reasoning that `omitempty` would handle the zero values. But you're right: `omitempty` applies to the inner fields, not the struct pointer itself... Reverted to the nil-guard approach. Thank you.

sleterrier:
Thanks for the thorough review @dm3ch, much appreciated! A few thoughts:

-  Actually, `shieldedInstanceConfig` was (very recently) added to `GCENodeClass` in #251. So Karpenter-provisioned VMs are covered. This PR specifically targets the template node pools, which are created with a minimal `NodeConfig` that doesn't satisfy org policy constraints we enforce.
- GCENodeClass-driven config sounds great. But from what I can tell the template node pools are created at controller startup by the `nodepooltemplate` controller, before any `GCENodeClass` resources exist. It seems like deriving template pool settings from `GCENodeClass` would require deferring pool creation until a `GCENodeClass` is reconciled. Happy to be corrected if I'm misreading the startup sequence.
- Agreed this isn't ideal long-term, especially for CMEK. But given the current architecture where template pools are created independently, controller-level options seemed like the pragmatic path and consistent with the existing `DEFAULT_NODEPOOL_SERVICE_ACCOUNT` pattern.

dm3ch:
It seems https://github.com/cloudpilot-ai/karpenter-provider-gcp/pull/251 have implemented that on GCENodeClass level as I proposed

If so I propose to close this PR

sleterrier:
Following up after investigating the broader architecture around your comment.

**Short term**, I only see two viable paths to unblock environments with org policy constraints:
1. This PR adds controller-level flags for CMEK and Shielded VM on template node pools, consistent with the existing `DEFAULT_NODEPOOL_SERVICE_ACCOUNT` pattern.
2. Or document pre-provisioning the `karpenter-default`/`karpenter-ubuntu`/etc. node pools out-of-band (e.g., via Terraform or CLI) with the required settings. The `ensureKarpenterNodePoolTemplate` logic already skips creation if the pool exists. This avoids new controller flags but pushes the responsibility to the operator.

**Longer term**, consolidating all config into `GCENodeClass` (as karpenter-aws does with `EC2NodeClass`) would mean eliminating the template node pool dependency entirely. From my understanding, that requires two things:

- Adding missing selector fields to `GCENodeClass`. At minimum `SubnetworkSelectorTerms` for network discovery, since networking is currently inherited from templates.
- Figuring out how to generate kube-env metadata without a GKE node pool. Today, the template node pools exist primarily to let GKE generate the bootstrap metadata (CA cert, API server endpoint, kubelet binary URLs, etc.). As far as I can tell, building this from scratch means reverse-engineering GKE's obscure internal node bootstrap mechanism, which is version-specific and not a public contract.

Happy to contribute to either path. For now, would merging this PR (or going with the out-of-band pre-provisioning approach) be acceptable as a stopgap?

dm3ch:
Hey @sleterrier, apologies for the confusion in my earlier comments — I initially misread the scope of this PR and incorrectly suggested that #251 already covers your use case.

  To be clear on what each PR actually does:
  - #251 (now merged) adds shieldedInstanceConfig to GCENodeClass, which controls Shielded VM settings on Karpenter-provisioned compute instances — i.e., actual workload VMs.
  - This PR targets the template node pools (karpenter-default, karpenter-ubuntu, etc.) that the nodepooltemplate controller auto-creates at startup. These are a different resource, and org policies like constraints/compute.requireShieldedVm enforce at node pool creation time too — so #251 doesn't help here.

  You're right that this is a real blocker for GCP projects with that org policy enforced, and I appreciate you investigating the architecture so thoroughly.

  That said, I have a concern about adding these as controller-level flags. The DEFAULT_NODEPOOL_SERVICE_ACCOUNT precedent exists for a practical reason (the controller needs an identity to operate), but growing this pattern with security settings feels like it creates technical debt: if we later move toward deriving
  template pool config from GCENodeClass (or eliminating the template pool dependency entirely), these Helm values would need to be removed.

  Your point about template pools being created before any GCENodeClass exists is valid and is exactly the architectural tension. There's no clean solution right now, which is why I'd prefer to agree on a direction before merging something we might need to undo.

  **What I'd like to propose:**
  1. Pause this PR for now.
  2. Open a separate issue describing the root problem (template node pool creation fails under compute.requireShieldedVm / gcp.restrictNonCmekServices), so we can discuss the long-term approach together with all maintainers and interested people.
  3. Use that issue to evaluate different directions. Currently I can imagine the following options (but I would be happy if you would find anything I haven't found):
    - Derive template pool config from GCENodeClass (requires deferring pool creation until at least one GCENodeClass is reconciled — complex but clean)
    - Use a single Karpenter node pool configured to satisfy the strictest org policies by default, since we override most behavior at instance creation time anyway
    - Eliminate the template pool dependency entirely by building bootstrap metadata from the GKE API directly — probably the ideal long-term solution, but requires understanding GKE's internal bootstrap contract, which you correctly flagged as non-trivial
  4. If the agreed-upon solution turns out to be a large effort, we can revisit a short-term fix — but with the long-term shape in mind, so it doesn't block future cleanup.

What do you think?

  P.S. The pre-provisioning workaround is worth documenting now regardless, since it's zero-code and immediately unblocks affected users (if you verified and it works).

  P.S.S. We've hit a similar architectural tension before (arm64 node provisioning), and I think it's worth taking a step back to agree on how we want to handle cases like this going forward before adding more operator-level options.

  P.S.S.S. Happy to create the issue myself if that's easier — just let me know, and thanks again for the detailed analysis of the bootstrap mechanism. That context will be useful regardless of which direction we choose.

sleterrier:
Sounds good! I'll open an issue with the problem description, the options we outlined, and the pre-provisioning workaround (which we've verified works in our environment with both `compute.requireShieldedVm` and `gcp.restrictNonCmekServices` for GCE enforced). Thanks for steering this in the right direction.

sleterrier:
Issue https://github.com/cloudpilot-ai/karpenter-provider-gcp/issues/255 created! please let me know if I forgot or messed up anything.

dm3ch:
@sleterrier What do you think about marking this PR as Draft or Closing it, until new long-term solution would be chosen?

sleterrier:
@dm3ch : of course, it makes sense we close this MR 👍  It seems very unlikely we would pick this solution long-term anyway.

PR #268, opened 2026-04-19
fix(instance): retry transient 503s in operation poll to prevent double-insertion

### LINKED ISSUE #267: bug: instance double-insertion when GCP returns 503 on operation poll
**What happened**:

When Karpenter creates a GCE instance, the `instances.insert` call succeeds and returns an operation name. If the subsequent `ZoneOperations.Get` poll returns a transient 503 (`Visibility check was unavailable`), the error is treated as fatal and propagated as a `retryableError`. The instance-type loop in `Create()` then advances to the next instance type, selects a **different zone**, and inserts a second VM — without checking whether the original insert actually completed.

Observed in an e2e run (SHA `091b543d`):

```
instance.go:414  "failed to wait for operation to be done"
  zone: us-central1-a
  error: "googleapi: Error 503: Visibility check was unavailable..."

instance.go:298  "Created instance"
  zone: us-central1-b   ← retry, different zone
```

Both lines share the same `reconcileID`.

The 503 from `ZoneOperations.Get` should be retried within the existing polling loop. Only a definitive error or context timeout should abort the poll. A second VM should never be created for the same NodeClaim.

Inject a 503 response into `ZoneOperations.Get` while the insert operation is in-flight. The polling loop exits immediately and `getOrCreateInstance` returns `retryable=true`, causing `Create()` to insert a duplicate instance.

Consequences of the bug:
1. VM in `us-central1-a` boots and joins the cluster with `providerID: gce://.../us-central1-a/...`
2. Karpenter records `NodeClaim.status.providerID = gce://.../us-central1-b/...` (the retry)
3. `providerID` mismatch → Node ↔ NodeClaim registration never completes
4. `karpenter.sh/unregistered=true:NoExecute` taint is never removed
5. Workload pods remain Pending indefinitely; orphan VM leaks cost

Root cause: `waitOperationDone` returns immediately on any error from `ZoneOperations.Get`, including transient 5xx responses. The fix is to continue the poll loop on transient errors (HTTP 5xx) and only exit on context cancellation/timeout or a non-transient error.

**Environment**:

- Karpenter-provider-gcp version: 0.2.1`091b543d`
- GKE version: 1.35
- Others: N/A

dm3ch:
#### What type of PR is this?

/kind bug

Adds retry on 5xx errors in `waitOperationDone`.

This allows us to prevent retry of the operation if we weren't able to retrieve the operation status just due to temporal server-side problems

For example, it fixes #267 where retry results in creation of a second VM, while the first one becomes an orphan.

Fixes #267

```release-note
Fix 5xx error handling during poll of operation status from GCP.
Fix instance double-insertion when GCP returns a transient 503 during operation polling.
Previously, a 503 from ZoneOperations.Get caused the retry loop to create a second VM in a different zone, leaving an orphaned VM and a broken NodeClaim providerID.
```

### DIFF
--- a/pkg/providers/instance/instance.go
+++ b/pkg/providers/instance/instance.go
@@ -124,6 +124,14 @@ func (p *DefaultProvider) waitOperationDone(ctx context.Context,
 			if waitCtx.Err() != nil {
 				return fmt.Errorf("waiting for operation %s: %w", operationName, waitCtx.Err())
 			}
+			if isTransientError(err) {
+				log.FromContext(ctx).V(1).Info("transient error polling operation, retrying",
+					"operation", operationName, "zone", zone, "error", err)
+				if tickErr := waitForNextTick(waitCtx, ticker); tickErr != nil {
+					return fmt.Errorf("waiting for operation %s: %w", operationName, tickErr)
+				}
+				continue
+			}
 			return fmt.Errorf("getting operation: %w", err)
 		}
 
@@ -167,6 +175,14 @@ func (p *DefaultProvider) handleZoneOperationError(ctx context.Context, op *comp
 		return e.Message
 	})
 	return fmt.Errorf("operation failed: %s", strings.Join(errorMsgs, "; "))
+}
+
+func isTransientError(err error) bool {
+	var apiErr *googleapi.Error
+	if errors.As(err, &apiErr) {
+		return apiErr.Code >= 500
+	}
+	return false
 }
 
 func isInsufficientCapacityError(operationError *compute.OperationErrorErrors) bool {
--- a/pkg/providers/instance/instance_test.go
+++ b/pkg/providers/instance/instance_test.go
@@ -19,8 +19,10 @@ package instance
 import (
 	"context"
 	"encoding/json"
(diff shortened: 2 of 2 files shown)


dm3ch:
### E2E Test Results

**Cluster:** `us-central1-f` | **Commit:** `6942e9f4` | **Total duration:** `5m 25s`

|---|---|---|---|---|
| Provisioning | Spot amd64 | ✅ pass | 1m 32s | |
| Provisioning | Spot arm64 | ✅ pass | 1m 21s | |
| Provisioning | On-Demand amd64 | ✅ pass | 1m 10s | |
| Provisioning | On-Demand arm64 | ✅ pass | 1m 15s | |
| Consolidation | consolidates on-demand nodes | ✅ pass | 2m 1s | |
| Drift | detects and replaces drifted nodes | ✅ pass | 2m 32s | |
| GarbageCollection | removes orphaned instances | ✅ pass | 1m 10s | |
| Networking | pod networking connectivity | ✅ pass | 1m 11s | |

**Overall: ✅ all pass**

thameezb (approved):

PR #270, opened 2026-04-20
feat(chart): add values.schema.json for Helm chart validation

dm3ch:
#### What type of PR is this?

/kind feature

Adds `charts/karpenter/values.schema.json` (JSON Schema draft 2020-12).

Benefits:
- **Input validation**: `helm install/upgrade` now rejects wrong types, invalid enum values (e.g. bad `logLevel` or `pullPolicy`), and out-of-range numbers before anything is deployed
- **Artifact Hub**: AH displays an interactive typed values reference when `has_values_schema` is true.

Required fields (`controller.settings.projectID`, `clusterName`, `clusterLocation`) are declared in the schema. All other fields have defaults matching `values.yaml`.

N/A

#### Docs and examples

```release-note
Adds values.schema.json to the Helm chart, enabling input validation on helm install/upgrade and typed values reference on Artifact Hub.
```

### DIFF
--- a/charts/karpenter/values.schema.json
+++ b/charts/karpenter/values.schema.json
@@ -1,0 +1,297 @@
+{
+  "$schema": "https://json-schema.org/draft/2020-12/schema",
+  "type": "object",
+  "additionalProperties": false,
+  "properties": {
+    "nameOverride": {
+      "type": "string",
+      "description": "Override the default chart name."
+    },
+    "fullnameOverride": {
+      "type": "string",
+      "description": "Override the full release name."
+    },
+    "logLevel": {
+      "type": "string",
+      "enum": ["debug", "info", "warn", "error"],
+      "description": "Global log level.",
+      "default": "info"
+    },
+    "logOutputPaths": {
+      "type": "array",
+      "items": { "type": "string" },
+      "description": "Log output paths.",
+      "default": ["stdout"]
+    },
+    "logErrorOutputPaths": {
+      "type": "array",
+      "items": { "type": "string" },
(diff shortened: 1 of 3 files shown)


charts/karpenter/values.schema.json
```diff
@@ -0,0 +1,297 @@
+{
```
thameezb, line 1: How do we ensure this is upto date? and was this generated? if so please add info to docs

dm3ch: The schema is hand-authored typically (but I used a LLM tool to generate that).
I've added a `$comment` field to the file itself to make this clear to future contributors.

`make chart-lint` invokes `helm lint` which checks our default `values.yaml` against schema and would fail if it found undocumented option or other violation of scheme.

So it would fail if values updated but schema changes haven't been reflected. 

Also it's being invoked in CI as part of `presubmit`.

thameezb: Can we add a script to automate the creation/generation of this or not worth the time/effort?

dm3ch: So far I don't know any tool able to do that. So I would propose to skip it for now.

In the future I have an idea to try to introduce some kind of AI review agent, and make a prompt to propose changes in docs / helm schemes.

UPD: Found https://github.com/losisin/helm-values-schema-json tool. But still don't have clear opinion if I want to add it. As it would mean that default values would be more authoritative and overwrite schema. In the mean time schema can have some validation rules for some values that are not set by default.

thameezb (approved):

Issue #239, opened 2026-04-04
feat: support reserved capacity type in getCapacityType (parity with karpenter-aws)


dm3ch:
**What would you like to be added**:

Support for a `reserved` capacity type in `getCapacityType` (`pkg/providers/instance/instance.go`), giving it highest priority in the selection order — mirroring the behaviour in karpenter-provider-aws.

karpenter-aws currently uses this priority hierarchy:

```
reserved > spot > on-demand
```

karpenter-gcp currently only handles:

```
spot > on-demand
```

The AWS implementation iterates over `[]string{karpv1.CapacityTypeReserved, karpv1.CapacityTypeSpot}` and returns the first type for which the NodeClaim allows it and a compatible available offering exists. GCP should do the same once a GCP-native reserved capacity concept is mapped to `karpv1.CapacityTypeReserved`.

**Why is this needed**:

GCP supports [node reservations](https://cloud.google.com/compute/docs/instances/reservations-overview). Users who have reservations today cannot express a preference for them via Karpenter; the capacity type selection logic simply ignores the reserved tier. Implementing this would allow reserved capacity to be consumed first, reducing waste and cost, consistent with how karpenter-aws works.

**Completion requirements**:

This enhancement requires the following artifacts:

- [ ] Design doc
- [ ] API change
- [ ] Docs update

PR #276, opened 2026-04-22
chore(deps): upgrade karpenter-core from v1.8.0 to v1.12.0

dm3ch:
#### What type of PR is this?

/kind cleanup

Upgrades `sigs.k8s.io/karpenter` from v1.8.0 to v1.12.0.

**Key changes in karpenter-core v1.9.0:**
- New `Gte`/`Lte` NodeSelectorOperators (inclusive bounds)
- New `DisableControllerWarmup` option: when false, controllers pre-populate caches before leader election, improving failover time
- `nodepoolhealth.State` refactored with proper mutex locking
- `RestrictedLabelDomains` extended to include `kubernetes.io` and `k8s.io`
- New `UnevaluatedNodePoolError` type in cloudprovider package

**Key changes in karpenter-core v1.10.0:**
- NodePool cost metric for enhanced cost visibility
- Static capacity controller: prevents modifying NodeClaims during NodePool deletion

**Key changes in karpenter-core v1.11.0:**
- New optional `NodeLifecycleHook` interface for cloud providers to gate node registration (not implemented here — tracked separately)
- `NodePool.spec.limits.nodes`: cap node count per NodePool independently of resource limits

**Key changes in karpenter-core v1.12.0:**
- `NodePool.spec.disruption.terminationGracePeriod`: configurable grace period before Karpenter force-deletes pods during disruption
- Re-sync of node labels after registration hooks complete

**GCP provider compat fixes (v1.9.0):**
- `image.go`: `NodeSelectorRequirementWithMinValues` no longer embeds `v1.NodeSelectorRequirement`; construct `corev1.NodeSelectorRequirement` directly from `item.Key/Operator/Values`
- `instance.go`: `Requirements.Labels()` removed; replaced with inline loop over `Requirements` filtering single-value `In` entries
- `instance.go`: Kubernetes label keys sanitized before being stamped as GCP VM labels

**CRD generation fix (v1.9.0):**
`hack/update-generated.sh` now copies karpenter-core CRDs from `vendor/sigs.k8s.io/karpenter/pkg/apis/crds/` instead of re-generating them via controller-gen.

**Helm chart additions:**
- `ReservedCapacity` feature gate (default false) - true in core, but we do not support this feature yet #239
- `DisableControllerWarmup` option (default true)
- `values.schema.json` restored with all feature gate entries

**New docs:**
- `docs/settings.md` — reference for all feature gates, controller options, and GCP-specific settings; includes new NodePool fields (`limits.nodes`, `terminationGracePeriod`)

Fixes #

#### Docs and examples

- [x] `docs/` updated — `docs/settings.md` added/updated

- This PR upgrades from v1.8.0 → v1.12.0 across 4 minor releases.
- Feature gates are unchanged across v1.9.0–v1.12.0 (same 5 gates, same defaults).
- CloudProvider interface is unchanged across v1.9.0–v1.12.0. No new required methods.
- `NodeLifecycleHook` (optional, v1.11) is not implemented here — tracked in a separate task.

```release-note
Upgraded karpenter-core from v1.8.0 to v1.12.0. New NodePool fields available: spec.limits.nodes (node count cap, v1.11) and spec.disruption.terminationGracePeriod (disruption grace period, v1.12). Helm chart now exposes ReservedCapacity feature gate (default on) and DisableControllerWarmup option. New docs/settings.md covers all configurable options.
```


sleterrier:
Hey @dm3ch, apologies for the intrusion on your PR but out of curiosity, why not upgrading to latest karpenter-core v1.11.1? I am sure there are very good reasons.

dm3ch:
> Hey @dm3ch, apologies for the intrusion on your PR but out of curiosity, why not upgrading to latest karpenter-core v1.11.1? I am sure there are very good reasons.

The community questions and activity is always welcome. :)

The reason is each minor version can change providers interfaces and have own hacks and bugs which should be addressed. For example, 1.8 and 1.9 upgrades both have interfaces changes, and 1.9 resulted in some problems with CRD generation. Also each minor version can implement new features, that have to be tested, adopted separately.

Due to this fact I initialy decided to upgrade 1 minor version at a time to scope changes and better control what changes between the versions.

But maybe you're right and this approach creates too many intermediate steps. Maybe as a next step would try to upgrade to a two minor versions in a row 1.9->1.11.1 :)

charts/karpenter/values.yaml
```diff
@@ -69,6 +69,12 @@ controller:
     # -- nodeRepair is ALPHA and is disabled by default.
     # When enabled, Karpenter replaces nodes that fail GKE Node Problem Detector health conditions.
     nodeRepair: false
+    # -- reservedCapacity enables scheduling to reserved/committed GCP capacity (default on).
+    reservedCapacity: true
```
thameezb: why is this set to true?

docs/settings.md
```diff
@@ -0,0 +1,44 @@
+# Settings Reference
+
+Karpenter for GCP is configured via environment variables passed to the controller pod. Most of these can be set through the Helm chart's `values.yaml`. The table below covers every configurable option.
+
+## Feature Gates
```
thameezb, line 5: The same question about us documenting upstream features, this will become an unmaintained copy-pasta list

dm3ch: Answered in: https://github.com/cloudpilot-ai/karpenter-provider-gcp/pull/277#discussion_r3140519074

Additional about this particular doc:
- It tracks our own list of supported feature gates, controller arguments and required parameters. It's imposible to know from karpenter.sh, cause it depends on the following factors: which version of karpenter-core we are depending; which our own uniq we have; which of them have been actually implemented/supported.

charts/karpenter/values.yaml
dm3ch: This matches the karpenter-core default — ReservedCapacity=true is set [upstream in this version](https://github.com/kubernetes-sigs/karpenter/blob/v1.9.0/pkg/operator/options/options.go#L172). We're explicitly exposing it in values.yaml so users can see and override it, rather than leaving it as a hidden default.

I don't see any particular reason to keep it false, as this feature is implemented fully on karpenter-core level, so if they decided that it's ready for use by default I have no objections to that.

dm3ch:
### E2E Test Results

**Cluster:** `us-central1-f` | **Commit:** `4742bbb3` | **Total duration:** `~12m`

|---|---|---|---|---|
| Consolidation | should consolidate an empty amd64 on-demand node | ✅ pass | 2m 29s | |
| Drift | should replace a drifted amd64 on-demand node | ✅ pass | 2m 19s | |
| Expiration | should replace an expired amd64 on-demand node | ✅ pass | 3m 59s | |
| GarbageCollection | should delete an orphaned VM after its NodeClaim is force-removed | ✅ pass | 3m 4s | |
| Networking | should provision a node with no external IP when enableExternalIPAccess is false | ✅ pass | 1m 45s | |
| NodeRepair | should replace a node whose KernelDeadlock condition has been True beyond the toleration | ✅ pass | 1m 38s | |
| Provisioning | COS / amd64 / on-demand | ✅ pass | 1m 36s | |
| Provisioning | COS / amd64 / spot | ✅ pass | 1m 31s | |
| Provisioning | COS / arm64 / on-demand | ✅ pass | 1m 37s | |
| Provisioning | COS / arm64 / spot | ✅ pass | 1m 26s | |
| Provisioning | Ubuntu / amd64 / on-demand | ✅ pass | 2m 14s | |
| Provisioning | Ubuntu / amd64 / spot | ✅ pass | 2m 29s | |
| Provisioning | Ubuntu / arm64 / on-demand | ❌ fail | 8s | Pre-existing: no target image found for arm64 Ubuntu node class |
| Provisioning | Ubuntu / arm64 / spot | ❌ fail | 8s | Pre-existing: no target image found for arm64 Ubuntu node class |

**Overall: ⚠️ 2 failures — 2 pre-existing** (Ubuntu arm64 image resolution fails on this cluster; unrelated to this PR)

charts/karpenter/values.yaml
thameezb: It's a new feature, in general I would assume these should be false until verified to work as expected within our implementation. 

Have the other flags been updated (to match upstream defaults)?

docs/settings.md
thameezb: As per other thread, if/when this becomes outdated we will need to relook at it

hack/update-generated.sh
```diff
@@ -2,9 +2,19 @@
...
-controller-gen crd paths=sigs.k8s.io/karpenter/pkg/apis/v1/... output:crd:dir=./charts/karpenter/crds
+
+# Copy karpenter-core CRDs from vendored pre-built files rather than re-generating
+# from source.  Re-generating via controller-gen would produce a truncated operator
+# enum for NodeSelectorRequirementWithMinValues.Operator ([Gte,Lte] only) because the
+# upstream marker (+kubebuilder:validation:Enum:=Gte;Lte) interacts differently with
+# every controller-gen version. The vendored files are the canonical artifacts shipped
+# by the karpenter-core team and always have the correct full enum.
+KARPENTER_CRD_DIR=vendor/sigs.k8s.io/karpenter/pkg/apis/crds
+cp "${KARPENTER_CRD_DIR}"/karpenter.sh_nodeclaims.yaml ./charts/karpenter/crds/
```
thameezb: When this list is changed, how will we know to do the same on ourside? This adds a manual step to the automated process without any notes regarding how to keep this maintained

charts/karpenter/values.yaml
dm3ch: Ok, good catch. I initially thought that `reservedCapacity` is about static nodepools size (but that's `StaticCapacity`). We do not really support `reservedCapacity` now, it's implementation is tracked in https://github.com/cloudpilot-ai/karpenter-provider-gcp/issues/239 .

I would set it to false and document that it's not supported yet

dm3ch: Disabled in - https://github.com/cloudpilot-ai/karpenter-provider-gcp/pull/276/changes/96bd5c1eab98bc523e528d78a65d8322dd7419a8

hack/update-generated.sh
dm3ch: Addressed that by copying all karpenter-core CRDs instead of copying the particular CRDs - https://github.com/cloudpilot-ai/karpenter-provider-gcp/pull/276/commits/6c1b10fe6390920f8435dffd66b715bacbee0530

thameezb (approved):

PR #263, opened 2026-04-19
feat: replace karpenter-managed bootstrap pools with pool discovery

dm3ch:
#### What type of PR is this?

/kind feature

#### What this PR does / why we need it:

Replaces the four hard-coded GKE node pools Karpenter created at startup to read bootstrap metadata (`karpenter-default`, `karpenter-ubuntu`, `karpenter-cos-arm64`, `karpenter-ubuntu-arm64`) with dynamic pool discovery:

- Discovers the first RUNNING/RUNNING\_WITH\_ERROR pool as bootstrap source (prefers `default-pool`, then alphabetical).
- `PatchKubeEnvForOSType` and `PatchKubeEnvForArch` adjust kube-env when the target node's OS or architecture differs from the source pool's.
- `karpenter-fallback` (zero-node COS pool, policy-hardened) created immediately only when no usable pool exists.
- `DEFAULT_NODEPOOL_TEMPLATE_NAME` / `--default-nodepool-template-name` pins the source pool explicitly.

Implements the design from [proposals/0001](proposals/0001-replace-karpenter-managed-bootstrap-pools.md). 

#### Which issue(s) this PR fixes:

Fixes #255 — pool creation was blocked on clusters enforcing `compute.requireShieldedVm` or `gcp.restrictNonCmekServices`; Karpenter now reuses an existing pool and only creates `karpenter-fallback` when strictly necessary.
Fixes #230 — pool creation failed on private-node clusters; `karpenter-fallback` mirrors the cluster's private-node setting automatically.

#### Docs and examples

- [x] `docs/` and `examples/` updated (or this PR does not affect user-facing behaviour, configuration, or APIs) - Related documentation PR: #329.
- [x] `MIGRATION.md` updated (or this PR does not require any migration steps)

#### Special notes for your reviewer:

- This PR was prepared with AI assistance (Claude Code). All changes have been reviewed and verified by the author.

<details open><summary><h3>Greptile Summary</h3></summary>

This PR replaces four hard-coded GKE node pools that Karpenter created at startup with dynamic pool discovery: it selects the first RUNNING/RUNNING_WITH_ERROR pool (preferring `default-pool`, then alphabetical), patches the selected pool's `kube-env` to adapt OS type and architecture when the target node differs from the source pool, and creates a zero-node `karpenter-fallback` pool only when no usable pool is found.

[...]

PR #277, opened 2026-04-22
docs: add node-repair and static-capacity pages; fix values schema for image.digest

dm3ch:
#### What type of PR is this?

/kind documentation
/kind bug

This PR adds the docs that should have accompanied those features, plus a schema fix for one other missing field.

**Schema fix (`charts/karpenter/values.schema.json`):**
- `image.digest` — present in `values.yaml` via #263 but absent from the schema, causing `helm lint` to reject it under `additionalProperties: false`

**Docs:**
- `docs/node-repair.md` — feature gate, monitored conditions, toleration durations, troubleshooting
- `docs/examples/static-capacity.md` — feature gate, warm-pool and static+dynamic NodePool patterns
- `docs/troubleshooting.md` — orphaned GCE instance GC section (#244)
- `docs/index.md` — Features section + links to all new pages

Fixes #

#### Docs and examples

- The `staticCapacity` ans `nodeRepair` schema fix originally in this PR was superseded by #287 and has been dropped from the diff.

```release-note
NONE
```

### DIFF
--- a/docs/examples/static-capacity.md
+++ b/docs/examples/static-capacity.md
@@ -4,6 +4,8 @@ A NodePool with `spec.replicas` set maintains a fixed number of nodes regardless
 
 This feature is disabled by default and requires the `staticCapacity` feature gate.
 
+> **Alpha:** This feature is controlled by the `staticCapacity` feature gate. Alpha features are off by default, may have known limitations, and their behaviour may change in future releases.
+
 ## Enabling the feature gate
 
 ```sh
--- a/docs/examples/static-capacity.md
+++ b/docs/examples/static-capacity.md
@@ -1,0 +1,115 @@
+# Static capacity (fixed node count)
+
+A NodePool with `spec.replicas` set maintains a fixed number of nodes regardless of pod demand. Karpenter provisions nodes up to that count at startup and replaces them if they are removed. Consolidation and `consolidateAfter` are ignored on static NodePools.
+
+This feature is disabled by default and requires the `staticCapacity` feature gate.
+
+## Enabling the feature gate
+
+```sh
+helm upgrade karpenter karpenter-provider-gcp/karpenter --install \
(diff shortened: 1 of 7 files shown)


docs/examples/static-capacity.md
```diff
@@ -0,0 +1,79 @@
+# Static capacity (fixed node count)
```
thameezb, line 1: Should we be documenting everything upstream feature in our docs? I would think not, else our docs becomes a slightly customised copy-pasta of upstream

dm3ch: Agreed the NodePool spec itself is generic — that part lives on karpenter.sh. What's missing there is whether this provider is on a karpenter-core version that ships it, and how to actually enable it on GCP (the staticCapacity Helm feature gate).

My rule: document upstream features when users can't answer "is this available in this provider right now? or How to use that?" from karpenter.sh alone — which is always true for new features, since upstream docs track karpenter-aws only (as we have different versioning not matching karpenter-aws). The doc here is that signal plus the one GCP-specific step (enabling the gate).

On maintenance: I agree this shouldn't become a burden. Long-term I'd like to explore auto-generating docs from CRDs and our own examples like promptless, that is being used by some of CNCF projects. And once we have proper docs hosting — embedding YAML snippets directly from `examples/` via mkdocs includes rather than duplicating them inline or making links to file from docs page.

This gap dates from an earlier karpenter-core upgrade that landed before we had doc coverage — caught it during the next upgrade and bundled it here.

dm3ch: So if I would describe my position very short: I believe that anything that requires additional explanation on top of karpenter.sh or configuration details should be documented.

thameezb: Lets see how it goes, if it becomes heavily out of state then we relook at it

thameezb (approved):

PR #260, opened 2026-04-16
docs(proposals): eliminate template node pool dependency

dm3ch:
#### What type of PR is this?

/kind documentation
/kind feature

#### What this PR does / why we need it:

Adds `docs/proposals/0001-eliminate-template-pool-dependency.md` — a KEP-style design proposal for eliminating Karpenter-managed GKE node pools from the bootstrap path (issue #255).

See the proposal doc for full context, design details, risks, and phased implementation plan.

#### Which issue(s) this PR fixes:

Related #255

#### Special notes for your reviewer:

This is a design doc only — no implementation code. Looking for feedback on:
1. Pool selection algorithm and scoring approach
2. Phase split — is Phase 3 a reasonable stopping point before full API reconstruction?
3. Risk assessments, especially Group 4 credential reuse
4. Phase 0 (e2e baseline) scope — anything missing from the test matrix?

- This PR was prepared with AI assistance (Claude Code). All changes have been reviewed and verified by the author.

#### Does this PR introduce a user-facing change?

```release-note
NONE
```

<details open><summary><h3>Greptile Summary</h3></summary>

- Adds a design proposal (KEP-style) for replacing Karpenter-managed GKE template node pools with a priority-based pool discovery algorithm, using `PatchKubeEnvForOSType` and `PatchKubeEnvForArch` for OS/arch patching and deferring pool creation only to a last-resort fallback path.
- The refresh/cache TTL is stated as **12 minutes** in the selection algorithm and code-changes table but as **5 minutes** in two separate risk sections, making the stale-credential and stale-template risk windows ambiguous.
- The `TPM_BOOTSTRAP_CERT`/`TPM_BOOTSTRAP_KEY` cross-pool reuse was listed as a hard PoC gate before implementation, but the Open Questions section has no \"Resolved\" entry for it even though a note states all phases were already implemented in PR #263.
</details>

<details><summary><h3>Confidence Score: 3/5</h3></summary>

Safe to merge as a documentation proposal, but two design-level inconsistencies should be resolved before implementation proceeds further.

[...]

PR #280, opened 2026-04-23
feat(imagefamily): resolve images from GKE catalog instead of template boot disks

dm3ch:
#### What type of PR is this?

/kind feature
/kind bug

Replace template-based image resolution with direct queries to the GCP image catalogs:
- COS images: queried from `gke-node-images` project, filtered to cluster K8s patch version
- Ubuntu images: queried from `ubuntu-os-gke-cloud` project, filtered to cluster K8s minor version

Previously Karpenter read the boot disk of a dedicated template pool (`karpenter-default` or `karpenter-ubuntu`) to discover the base image, then applied regex substitutions for arm64/GPU variants. This approach had two problems:

1. **arm64-ubuntu was broken**: the old regex assumed the `ubuntu-gke-2204` naming scheme (no arch suffix on amd64 images). The current `ubuntu-gke-2404` images carry explicit `-amd64-`/`-arm64-` suffixes, so the regex produced wrong image names.
2. **Template dependency**: image resolution failed if the template pool was not yet created, forcing a strict startup ordering dependency.

The new approach queries the catalog directly, picks the newest non-deprecated amd64 image, and derives arm64/GPU variants with a simple string substitution. Image resolution is now independent of template pool availability.

Fixes #

#### Docs and examples

- 12 new unit tests cover catalog filtering, usability checks, and arch-variant derivation.
- This PR was split from #263 and is part of the implementation of #260 (getting rid of Karpenter-managed GKE NodePools)

```release-note
Image resolution for COS and Ubuntu node images now queries GCP image catalogs (gke-node-images, ubuntu-os-gke-cloud) directly instead of reading template pool boot disks. arm64-Ubuntu provisioning is fixed as a result.
```


charts/karpenter/values.schema.json
```diff
@@ -221,14 +221,14 @@
               "description": "Enable NodeOverlay (alpha). Allows node overlay to influence scheduling decisions.",
               "default": false
             },
-            "nodeRepair": {
+            "staticCapacity": {
```
thameezb: Unrelated changes

dm3ch: Fixed, was a rebase conflict issue.

pkg/providers/imagefamily/ubuntu.go
```diff
@@ -18,39 +18,126 @@ package imagefamily
...
+// ubuntu-gke-2404 images use explicit arch in the name (e.g. ubuntu-gke-2404-1-35-amd64-v20260416),
+// unlike the older ubuntu-gke-2204 series which had no arch suffix on amd64 images.
+func (u *Ubuntu) resolveLatestUbuntuImage(ctx context.Context) (string, error) {
+	filter := u.buildImageFilter(ctx)
+
+	// GCP does not support Filter + OrderBy together, and there can be thousands of
+	// ubuntu-gke-2404 images (most deprecated). We page through all of them,
+	// collect non-deprecated amd64 candidates, then sort in code.
+	var candidates []*compute.Image
+	err := u.computeService.Images.List(ubuntuGKEImageProject).
```
thameezb, line 77: This seems quite memory intensive...is it?

pkg/providers/imagefamily/ubuntu.go
```diff
@@ -18,39 +18,126 @@ package imagefamily
...
 	if err != nil {
-		log.FromContext(ctx).Error(err, "Failed to get source image")
+		log.FromContext(ctx).Error(err, "failed to resolve Ubuntu GKE image from catalog")
 		return Images{}, err
 	}
 
 	if version != "latest" {
-		re := regexp.MustCompile(`v\d+$`)
-		sourceImage = re.ReplaceAllString(sourceImage, version)
+		re := regexp.MustCompile(`-v\d+(-|$)`)
```
thameezb: Could probably shift this to be a const so that we don't recompile it on every call of this function

thameezb (approved):
optimisation comments

pkg/providers/imagefamily/ubuntu.go
dm3ch: Good catch - addressed in https://github.com/cloudpilot-ai/karpenter-provider-gcp/pull/280/changes/9e6031742d8f0e3a1e06a3af159d870b2c92632e by moving it to package level var

pkg/providers/imagefamily/ubuntu.go
dm3ch: Haven't measured, don't think so, but it can become a problem if a lot of images matching prefix.

Paginated list shouldn't be the memory intensive itself. But candidates slice can in case we have really big amount of candidates. https://github.com/cloudpilot-ai/karpenter-provider-gcp/pull/280/changes/9e6031742d8f0e3a1e06a3af159d870b2c92632e - replaced `candidates` slice with storing only a `best` match

dm3ch:
### E2E Test Results

**Cluster:** `europe-west4-a` | **Commit:** `9e603174` | **Total duration:** `~20m` (parallel)

| Suite | Spec | Result | Duration |
|---|---|---|---|
| Consolidation | should consolidate an empty amd64 on-demand node | ✅ pass | 3m 7s |
| Drift | should replace a drifted amd64 on-demand node | ✅ pass | 3m 47s |
| Expiration | should replace an expired amd64 on-demand node | ✅ pass | 5m 38s |
| GC | should delete an orphaned VM after its NodeClaim is force-removed | ✅ pass | 2m 20s |
| Networking | should provision a node with no external IP when enableExternalIPAccess is false | ✅ pass | 2m 3s |
| Provisioning | COS / amd64 / on-demand | ✅ pass | 2m 11s |
| Provisioning | COS / arm64 / on-demand | ✅ pass | 2m 4s |
| Provisioning | COS / amd64 / spot | ✅ pass | 1m 49s |
| Provisioning | COS / arm64 / spot | ✅ pass | 1m 39s |
| Provisioning | Ubuntu / amd64 / on-demand | ✅ pass | 2m 53s |
| Provisioning | Ubuntu / arm64 / on-demand | ✅ pass | 2m 47s |
| Provisioning | Ubuntu / amd64 / spot | ✅ pass | 2m 57s |
| Provisioning | Ubuntu / arm64 / spot | ✅ pass | 2m 57s |
| Repair | should replace a node whose KernelDeadlock condition has been True beyond the toleration | ✅ pass | 3m 3s |

**Overall: ✅ all pass** — 14/14 specs, no panics in Karpenter logs.

PR #285, opened 2026-04-24
feat(chart): add karpenter-crd Helm chart for independent CRD lifecycle

dm3ch:
#### What type of PR is this?

/kind feature

Adds a separate `karpenter-crd` Helm chart that manages CRD lifecycle independently from the controller chart. This is the same pattern used by karpenter-provider-aws.

**Why separate charts?**
- `helm upgrade karpenter` (controller chart) does not upgrade CRDs in the `crds/` directory — Helm only applies those on `helm install`
- Putting CRDs in a separate chart's `templates/` directory ensures `helm upgrade karpenter-crd` keeps CRDs up to date
- Making `karpenter-crd` a Helm dependency of `karpenter` would cause `helm uninstall karpenter` to cascade-delete all CRDs, destroying all NodePool and NodeClaim resources

**What's included:**
- `charts/karpenter-crd/` — new chart with CRDs in `templates/`, `additionalAnnotations` value (for ArgoCD `sync-options` etc.), JSON schema validation
- `hack/mutation/crd_annotations.sh` — injects the `additionalAnnotations` Helm block into each CRD template (matching AWS's approach)
- `hack/update-generated.sh` — syncs CRDs into `karpenter-crd` templates on every `make update`; cleans templates dir first to remove stale CRDs
- `Makefile` — `chart-lint` and `update` targets cover both charts
- Release workflow — packages and publishes both charts; restore step uses awk instead of Python to parse the Helm index
- e2e deploy script — installs `karpenter-crd` before the controller
- Docs — two-chart install pattern in `installation.md`; dedicated `upgrading.md`

Fixes #

#### Docs and examples

- CRD sync follows the same awk-based pattern as [karpenter-provider-aws](https://github.com/aws/karpenter-provider-aws/blob/main/hack/mutation/crd_annotations.sh)

```release-note
Added `karpenter-crd` Helm chart for independent CRD lifecycle management. Install CRDs separately with `helm upgrade --install karpenter-crd karpenter-provider-gcp/karpenter-crd` before upgrading the controller chart.
```

### DIFF
--- a/.github/workflows/release-build.yaml
+++ b/.github/workflows/release-build.yaml
@@ -104,26 +104,29 @@ jobs:
 
           helm repo add karpenter-gcp "${PAGES_URL}" --force-update
 
-          # Parse "chart_name version" pairs from the Helm index YAML using awk.
-          # Chart names appear at 2-space indent (e.g. "  karpenter:"); versions at 4-space indent.
-          awk '/^  [a-z].*:$/ { chart=$1; sub(/:$/,"",chart) } /^    version:/ { print chart " " $2 }' \
-            /tmp/merge-index.yaml > /tmp/chart-versions.txt
+          for CHART in karpenter karpenter-crd; do
+            VERSIONS=$(helm search repo "karpenter-gcp/${CHART}" --versions -o json 2>/dev/null | jq -r '.[].version' || true)
+            [ -z "${VERSIONS}" ] && continue
 
-          while read CHART VERSION; do
-            TARBALL="charts/${CHART}-${VERSION}.tgz"
-            if [ ! -f "${TARBALL}" ]; then
-              echo "Pulling ${CHART} ${VERSION}..."
-              helm pull "karpenter-gcp/${CHART}" --version "${VERSION}" --destination charts/ || \
-                echo "WARNING: could not pull ${CHART} ${VERSION}, skipping"
+            echo "${VERSIONS}" | while read VERSION; do
+              if [ ! -f "charts/${CHART}-${VERSION}.tgz" ]; then
+                echo "Pulling ${CHART} ${VERSION}..."
+                helm pull "karpenter-gcp/${CHART}" --version "${VERSION}" --destination charts/ || \
+                  echo "WARNING: could not pull ${CHART} ${VERSION}, skipping"
+              fi
+            done
+
+            # Fail rather than deploy an incomplete repo
+            ACTUAL=0
+            for V in ${VERSIONS}; do
+              [ -f "charts/${CHART}-${V}.tgz" ] && ACTUAL=$((ACTUAL + 1))
+            done
+            EXPECTED=$(echo "${VERSIONS}" | grep -c .)
+            if [ "${ACTUAL}" -lt "${EXPECTED}" ]; then
+              echo "ERROR: Expected ${EXPECTED} ${CHART} archives but only have ${ACTUAL}. Aborting."
+              exit 1
             fi
-          done < /tmp/chart-versions.txt
-
-          EXPECTED=$(wc -l < /tmp/chart-versions.txt | tr -d ' ')
-          ACTUAL=$(ls charts/karpenter*.tgz 2>/dev/null | wc -l | tr -d ' ')
-          if [ "${ACTUAL}" -lt "${EXPECTED}" ]; then
-            echo "ERROR: Expected ${EXPECTED} chart archives but only have ${ACTUAL}. Aborting."
-            exit 1
-          fi
+          done
 
       - name: Setup Pages
         uses: actions/configure-pages@v5
--- a/.github/workflows/release-build.yaml
+++ b/.github/workflows/release-build.yaml
@@ -81,8 +81,10 @@ jobs:
             exit 1
           fi
 
-      - name: Package chart
-        run: helm package charts/karpenter --version "${{ env.tag_semver }}" --app-version "${{ env.tag }}" -d charts/
+      - name: Package charts
+        run: |
+          helm package charts/karpenter     --version "${{ env.tag_semver }}" --app-version "${{ env.tag }}" -d charts/
+          helm package charts/karpenter-crd --version "${{ env.tag_semver }}" --app-version "${{ env.tag }}" -d charts/
 
       - name: Update Helm repo index
         run: |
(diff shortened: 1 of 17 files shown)


sleterrier:
Thanks for picking this up — sorry for the duplicate effort with #288 (now closed). Three things from my independent implementation that might be worth folding in:

### 1. Adoption snippet for existing installs

Existing in-cluster CRDs were created by Helm as install-only resources of the previous `karpenter` release, so they're not currently owned by any Helm release. `helm install karpenter-crd` then errors with `cannot patch <crd> ... already exists`. The fix is a one-time per-cluster `kubectl annotate/label` to make the CRDs look helm-managed under the new release; Helm 3 then adopts them on install.

```bash
for crd in gcenodeclasses.karpenter.k8s.gcp \
           nodeclaims.karpenter.sh \
           nodeoverlays.karpenter.sh \
           nodepools.karpenter.sh; do
  kubectl annotate crd $crd \
    meta.helm.sh/release-name=karpenter-crd \
    meta.helm.sh/release-namespace=karpenter-system \
    --overwrite
  kubectl label crd $crd app.kubernetes.io/managed-by=Helm --overwrite
done
```

Without this, every existing cluster blocks at the first `helm install karpenter-crd`. Probably worth a section in `upgrading.md` ahead of the regular upgrade procedure.

### 2. `helm search repo` instead of awk-parsing the merge index

In `.github/workflows/release-build.yaml`, the awk pass over `/tmp/merge-index.yaml` works but is fragile to YAML format changes. I think we can get the same data from helm directly:

```bash
VERSIONS=$(helm search repo "karpenter-gcp/${CHART}" --versions -o json 2>/dev/null | jq -r '.[].version' || true)
[ -z "${VERSIONS}" ] && continue
```

Works with the existing `helm repo add karpenter-gcp --force-update`. No YAML grovelling, returns empty cleanly when the chart doesn't exist yet (first-release case for `karpenter-crd`).

### 3. Remove `charts/karpenter/crds/`

With the new install order documented (`karpenter-crd` first, then `karpenter`), keeping the CRDs in `charts/karpenter/crds/` is redundant and creates a conflict: `helm install karpenter` would install the CRDs as install-only resources of the `karpenter` release, and a subsequent `helm install karpenter-crd` would then collide with `cannot patch <crd> ... already exists`. Same failure mode #1 fixes for old installs, just on a fresh cluster.

`charts/karpenter-crd/templates/` should be the single source of truth for the four CRDs.

docs/getting-started/upgrading.md
```diff
@@ -0,0 +1,30 @@
...
+# 2. Upgrade the controller
+helm upgrade karpenter karpenter-provider-gcp/karpenter \
+  --namespace karpenter-system
+```
+
+If you installed with the single-chart pattern before `karpenter-crd` was available, migrate by installing the CRD chart once:
+
+```sh
+helm upgrade --install karpenter-crd karpenter-provider-gcp/karpenter-crd
+```
```
sleterrier: Adoption snippet for existing installs.

````suggestion
If you installed with the single-chart pattern before `karpenter-crd` was available, the in-cluster CRDs are not currently owned by any Helm release. Adopt them into the new `karpenter-crd` release first, so Helm doesn't error with `cannot patch <crd> ... already exists`:

```bash
for crd in gcenodeclasses.karpenter.k8s.gcp \
           nodeclaims.karpenter.sh \
           nodeoverlays.karpenter.sh \
           nodepools.karpenter.sh; do
  kubectl annotate crd $crd \
    meta.helm.sh/release-name=karpenter-crd \
    meta.helm.sh/release-namespace=karpenter-system \
    --overwrite
  kubectl label crd $crd app.kubernetes.io/managed-by=Helm --overwrite
done
```

Then install the CRD chart:

```sh
helm upgrade --install karpenter-crd karpenter-provider-gcp/karpenter-crd
```
````

File: .github/workflows/release-build.yaml [hunk 2]
```diff
@@ -98,22 +100,28 @@ jobs:
...
-          # Fail rather than deploy an incomplete repo
-          EXPECTED=$(echo "${VERSIONS}" | grep -c .)
-          ACTUAL=$(ls charts/karpenter-*.tgz 2>/dev/null | wc -l | tr -d ' ')
+          EXPECTED=$(wc -l < /tmp/chart-versions.txt | tr -d ' ')
+          ACTUAL=$(ls charts/karpenter*.tgz 2>/dev/null | wc -l | tr -d ' ')
           if [ "${ACTUAL}" -lt "${EXPECTED}" ]; then
-            echo "ERROR: Expected ${EXPECTED} chart versions but only have ${ACTUAL}. Aborting."
+            echo "ERROR: Expected ${EXPECTED} chart archives but only have ${ACTUAL}. Aborting."
             exit 1
           fi
```
sleterrier: Replace the awk-into-tempfile + while-loop + EXPECTED/ACTUAL check with a per-chart loop using `helm search repo`. Same end result, no YAML grovel.

```suggestion
          for CHART in karpenter karpenter-crd; do
            VERSIONS=$(helm search repo "karpenter-gcp/${CHART}" --versions -o json 2>/dev/null | jq -r '.[].version' || true)
            [ -z "${VERSIONS}" ] && continue

            echo "${VERSIONS}" | while read VERSION; do
              if [ ! -f "charts/${CHART}-${VERSION}.tgz" ]; then
                echo "Pulling ${CHART} ${VERSION}..."
                helm pull "karpenter-gcp/${CHART}" --version "${VERSION}" --destination charts/ || \
                  echo "WARNING: could not pull ${CHART} ${VERSION}, skipping"
              fi
            done

            # Fail rather than deploy an incomplete repo
            ACTUAL=0
            for V in ${VERSIONS}; do
              [ -f "charts/${CHART}-${V}.tgz" ] && ACTUAL=$((ACTUAL + 1))
            done
            EXPECTED=$(echo "${VERSIONS}" | grep -c .)
            if [ "${ACTUAL}" -lt "${EXPECTED}" ]; then
              echo "ERROR: Expected ${EXPECTED} ${CHART} archives but only have ${ACTUAL}. Aborting."
              exit 1
            fi
          done
```

sleterrier (commented):
Inline follow-ups for suggestions 1 and 2 from the previous comment.

dm3ch:
@sleterrier 1 and 2 agree, would rework that.

3 are you sure we want to drop CRDs from `karpenter` chart. I agree that it's better to use separate charts installation, but for example karpenter-aws is still keeping crds in base chart, to allow easy install. Also some tooling like flux is capable of updating CRDs without a separate CRD chart - https://github.com/aws/karpenter-provider-aws/tree/main/charts/karpenter/crds (`CreateReplace` policy)

sleterrier:
@dm3ch fair point. Flux's CreateReplace and the AWS pattern both make a reasonable case. Leaving people to pick their poison.

dm3ch:
@sleterrier For some reason, apply on github failed. Applied on https://github.com/cloudpilot-ai/karpenter-provider-gcp/pull/285/changes/48efa578130ee107cf6234e71f16f014c1246dd0

sleterrier (approved):
Looks good to me 🚢

dm3ch:
Failed CI due to main - fix https://github.com/cloudpilot-ai/karpenter-provider-gcp/pull/299

thameezb (approved):

PR #289, opened 2026-04-26
test(drift): fix nil nodeClass panic; add unit tests and e2e NodeClassDrift coverage

dm3ch:
#### What type of PR is this?

/kind failing-test
/kind bug

These three changes are bundled intentionally: the nil fix was discovered while authoring the e2e test, and the unit tests directly cover the same code path guarded by the fix.

**1. Nil safety fix** (`pkg/cloudprovider/cloudprovider.go`)

`resolveNodeClassFromNodePool` returns `nil, nil` when the referenced `GCENodeClass` has a `DeletionTimestamp` (i.e. is being deleted). Before this fix, two callers were not guarding against that nil return:

- `GetInstanceTypes` — passed `nil` to `instanceTypeProvider.List`, causing a nil pointer dereference
- `IsDrifted` — passed `nil` to `isNodeClassDrifted`, causing a nil pointer dereference

Both methods now return early (`nil, nil` / `"", nil`) when `nodeClass == nil`, matching the intent of the existing comment: a terminating NodeClass is treated as non-drifted and has no instance types.

**When does this happen?** During normal NodePool churn: Karpenter deletes a NodeClass while an informer event for an associated NodeClaim is still in the queue. The reconciler calls `IsDrifted` or `GetInstanceTypes` on the now-terminating NodeClass, hits the nil path, and panics. This is most likely to surface under load or during e2e teardown where many resources are deleted concurrently.

**2. Unit tests for drift detection** (`pkg/cloudprovider/drift_test.go` — new file)

- `TestAreStaticFieldsDrifted` — 8 table-driven cases covering all annotation/version combinations (hash match, mismatch, missing annotations, version mismatch)
- `TestIsImageDrifted` — 4 table-driven cases covering ImageID suffix matching (match, multi-image, no match, empty)
- `TestGCENodeClassHashCoversAllMutableFields` — reflection-based: iterates every exported `GCENodeClassSpec` field and verifies each non-ignored field changes the hash; confirms `ImageSelectorTerms` (tagged `hash:"ignore"`) does not

**3. NodeClassDrift e2e test** (`test/suites/drift/drift_test.go`)

Provisions a node, patches `spec.metadata` on the `GCENodeClass` (a mutable field that participates in the hash), and verifies the node is replaced via `NodeClassDrift`. This is the simplest mutation that reliably triggers hash-based drift detection without needing a new NodePool or image.

The e2e test is what exposed the nil crash: during test teardown the NodeClass is deleted while the drift controller is still reconciling, triggering the nil dereference described above.

Fixes #

#### Docs and examples

- The nil fix and tests are intentionally in one PR: the crash was found while writing the e2e test, and the unit tests cover the same code path guarded by the fix. Splitting them would create a PR ordering dependency with no review benefit.

```release-note
NONE
```

### DIFF
--- a/pkg/cloudprovider/cloudprovider.go
+++ b/pkg/cloudprovider/cloudprovider.go
@@ -202,6 +202,9 @@ func (c *CloudProvider) GetInstanceTypes(ctx context.Context, nodePool *karpv1.N
 		log.FromContext(ctx).Error(gcperrors.ErrResolveNodeClass, "nodePool", nodePool)
 		return nil, fmt.Errorf("resolving node class, %w", err)
 	}
+	if nodeClass == nil {
+		return nil, nil
+	}
 	// TODO, break this coupling
 	instanceTypes, err := c.instanceTypeProvider.List(ctx, nodeClass)
 	if err != nil {
@@ -248,6 +251,9 @@ func (c *CloudProvider) IsDrifted(ctx context.Context, nodeClaim *karpv1.NodeCla
 			c.recorder.Publish(cloudproviderevents.NodePoolFailedToResolveNodeClass(nodePool))
 		}
 		return "", client.IgnoreNotFound(fmt.Errorf("resolving node class, %w", err))
+	}
+	if nodeClass == nil {
+		return "", nil
 	}
 	return c.isNodeClassDrifted(ctx, nodeClaim, nodePool, nodeClass), nil
 }
--- a/pkg/cloudprovider/cloudprovider.go
+++ b/pkg/cloudprovider/cloudprovider.go
@@ -46,6 +46,11 @@ import (
 
 const CloudProviderName = "gcp"
 
+// errNodeClassTerminating is returned by resolveNodeClassFromNodePool when the
+// referenced GCENodeClass has a non-zero DeletionTimestamp. Callers treat it as
+// "no NodeClass available" (equivalent to NotFound) without logging an error.
+var errNodeClassTerminating = stderrors.New("GCENodeClass is terminating")
+
 var _ cloudprovider.CloudProvider = (*CloudProvider)(nil)
 
 type CloudProvider struct {
@@ -195,15 +200,15 @@ func (c *CloudProvider) LivenessProbe(req *http.Request) error {
 func (c *CloudProvider) GetInstanceTypes(ctx context.Context, nodePool *karpv1.NodePool) ([]*cloudprovider.InstanceType, error) {
 	nodeClass, err := c.resolveNodeClassFromNodePool(ctx, nodePool)
 	if err != nil {
+		if stderrors.Is(err, errNodeClassTerminating) {
+			return nil, nil
+		}
 		if errors.IsNotFound(err) {
 			// If we can't resolve the NodeClass, then it's impossible for us to resolve the instance types
 			c.recorder.Publish(cloudproviderevents.NodePoolFailedToResolveNodeClass(nodePool))
 		}
 		log.FromContext(ctx).Error(gcperrors.ErrResolveNodeClass, "nodePool", nodePool)
 		return nil, fmt.Errorf("resolving node class, %w", err)
-	}
-	if nodeClass == nil {
-		return nil, nil
 	}
 	// TODO, break this coupling
 	instanceTypes, err := c.instanceTypeProvider.List(ctx, nodeClass)
@@ -219,9 +224,7 @@ func (c *CloudProvider) resolveNodeClassFromNodePool(ctx context.Context, nodePo
 		return nil, err
 	}
 	if !nodeClass.DeletionTimestamp.IsZero() {
-		// For the purposes of NodeClass CloudProvider resolution, we treat deleting NodeClasses as NotFound,
-		// but we return a different error message to be clearer to users
-		return nil, nil // return nil, newTerminatingNodeClassError(nodeClass.Name)
+		return nil, errNodeClassTerminating
 	}
 	return nodeClass, nil
 }
@@ -247,13 +250,13 @@ func (c *CloudProvider) IsDrifted(ctx context.Context, nodeClaim *karpv1.NodeCla
 	}
 	nodeClass, err := c.resolveNodeClassFromNodePool(ctx, nodePool)
 	if err != nil {
+		if stderrors.Is(err, errNodeClassTerminating) {
+			return "", nil
+		}
 		if errors.IsNotFound(err) {
 			c.recorder.Publish(cloudproviderevents.NodePoolFailedToResolveNodeClass(nodePool))
 		}
 		return "", client.IgnoreNotFound(fmt.Errorf("resolving node class, %w", err))
-	}
-	if nodeClass == nil {
-		return "", nil
 	}
 	return c.isNodeClassDrifted(ctx, nodeClaim, nodePool, nodeClass), nil
 }
--- a/test/pkg/environment/helpers.go
+++ b/test/pkg/environment/helpers.go
@@ -63,7 +63,7 @@ type TestCase struct {
 	InstanceTypes []string
 	// ImageFamily selects the OS image family for the NodeClass.
 	// Defaults to ContainerOptimizedOS when empty.
-	ImageFamily         string
+	ImageFamily string
 	ConsolidationPolicy string // defaults to WhenEmptyOrUnderutilized when empty
 }
 
--- a/test/pkg/environment/helpers.go
+++ b/test/pkg/environment/helpers.go
@@ -18,6 +18,7 @@ package environment
 
 import (
 	"context"
+	"encoding/json"
 	"errors"
 	"fmt"
 	"math/rand/v2"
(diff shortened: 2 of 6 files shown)


dm3ch:
### E2E Test Results

**Cluster:** \`us-central1-f\` | **Commit:** \`5f0b3356\` (binary) / \`10c7f776\` (branch HEAD) | **Total duration:** \`~30m\`

|---|---|---|---|---|
| Consolidation | should consolidate an empty amd64 on-demand node | ❌ fail | ~3m | GCP on-demand ICE (n2-standard-2/4, all zones) |
| Drift | should replace a drifted amd64 on-demand node | ❌ fail | 3m | GCP on-demand ICE — pre-existing infra |
| Drift | should replace a node when GCENodeClass metadata changes | ❌ fail | 3m | GCP on-demand ICE — pre-existing infra |
| Expiration | should expire an amd64 on-demand node | ❌ fail | ~3m | GCP on-demand ICE — pre-existing infra |
| GC | should garbage collect a leaked instance | ❌ fail | ~3m | GCP on-demand ICE — pre-existing infra |
| Networking | should provision a node with no external IP | ❌ fail | 10m | GCP on-demand ICE + cluster sync delay |
| Provisioning | COS / amd64 / on-demand | ❌ fail | 3m | GCP on-demand ICE — pre-existing infra |
| Provisioning | COS / arm64 / on-demand | ❌ fail | 3m | GCP on-demand ICE — pre-existing infra |
| Provisioning | COS / amd64 / spot | ❌ fail | 3m | GCP on-demand ICE — pre-existing infra |
| Provisioning | COS / arm64 / spot | ❌ fail | 3m | GCP on-demand ICE — pre-existing infra |
| Provisioning | Ubuntu / amd64 / on-demand | ❌ fail | 3m | GCP on-demand ICE — pre-existing infra |
| Provisioning | Ubuntu / arm64 / on-demand | ❌ fail | <1m | Known pre-existing: no arm64 Ubuntu image |
| Provisioning | Ubuntu / amd64 / spot | ✅ pass | 2m 41s | Spot capacity available |
| Provisioning | Ubuntu / arm64 / spot | ❌ fail | <1m | Known pre-existing: no arm64 Ubuntu image |
| Repair | should replace a node whose KernelDeadlock condition has been True beyond the toleration | ❌ fail | 13m | GCP on-demand ICE across all zones |

**Overall: ⚠️ 1 pass, 14 fail — all failures are pre-existing infrastructure constraints, not regressions from this PR**

**Root cause of failures (not code-related):**

1. **GCP on-demand capacity constraint**: `n2-standard-2` and `n2-standard-4` return ICE (Insufficient Capacity Error) across all us-central1 zones (a/b/c/f). Karpenter correctly creates NodeClaims and attempts provisioning, but GCP has no on-demand capacity available. After the first failed attempt, both instance types are cached as unavailable for 30 min, causing subsequent scheduling attempts to filter out all instance types. Spot capacity remains unaffected (Ubuntu/amd64/spot passed).

2. **Parallel-load cluster sync delay** (karpenter-core v1.12): 7 suites starting simultaneously triggers `cluster is waiting on sync for extended duration` on some NodePools for up to 9+ minutes, preventing scheduling from proceeding before tests time out.

3. **Ubuntu arm64** (pre-existing, unrelated): `no target image found` for arm64 Ubuntu — not a regression.

**PR changes are correct**: The nil guards added in this PR (`GetInstanceTypes`/`IsDrifted` returning `nil, nil` when NodeClass has DeletionTimestamp) were confirmed NOT to be involved in any of these failures. Karpenter successfully discovered pods, created NodeClaims, and reached GCP APIs — it was GCP that returned ICE, not Karpenter failing to schedule.

test/pkg/environment/helpers.go
```diff
@@ -630,6 +632,24 @@ func (e *Environment) WaitForNodeClassReady(ctx context.Context, name string) {
 	e.waitForReadyCondition(ctx, gceNodeClassGVR, name, NodeClassReadyTimeout)
 }
 
+// AddNodeClassMetadataEntry adds a single key/value pair to spec.metadata of the
+// named GCENodeClass using a merge patch. This is the simplest way to trigger
+// NodeClassDrift in tests — the hash controller detects the spec change, updates
+// the NodeClass hash annotation, and the disruption controller marks affected
+// NodeClaims as drifted.
+func (e *Environment) AddNodeClassMetadataEntry(ctx context.Context, name, key, value string) {
```
thameezb, line 640: If its just used by a test, does it need to be a public function?

pkg/cloudprovider/cloudprovider.go
```diff
@@ -202,6 +202,9 @@ func (c *CloudProvider) GetInstanceTypes(ctx context.Context, nodePool *karpv1.N
 		log.FromContext(ctx).Error(gcperrors.ErrResolveNodeClass, "nodePool", nodePool)
 		return nil, fmt.Errorf("resolving node class, %w", err)
 	}
+	if nodeClass == nil {
```
thameezb: When would this be nil but error not be nil?

pkg/cloudprovider/cloudprovider.go
```diff
@@ -249,6 +252,9 @@ func (c *CloudProvider) IsDrifted(ctx context.Context, nodeClaim *karpv1.NodeCla
 		}
 		return "", client.IgnoreNotFound(fmt.Errorf("resolving node class, %w", err))
 	}
+	if nodeClass == nil {
```
thameezb: As above

test/pkg/environment/helpers.go
dm3ch: This package `test/pkg/environment` is created only to store e2e helper functions.

As each e2e suite is a separate package we have a cross-package usage here and defenetlly need a public function

pkg/cloudprovider/cloudprovider.go
dm3ch: As it was said in a PR description. This check is to avoid panic of karpenter when NodeClass is being deleted, but some resource is still referencing it (as far as I understand deletion of dependant resources happens afterwards).

In https://github.com/cloudpilot-ai/karpenter-provider-gcp/pull/289/commits/0941c313fcbe214e086af6a45f00fb599694bef1 I introduced new `errNodeClassTerminating` and now checking for this particular error type. It should be much better readable

pkg/cloudprovider/cloudprovider.go
dm3ch: https://github.com/cloudpilot-ai/karpenter-provider-gcp/pull/289#discussion_r3161641999

pkg/cloudprovider/cloudprovider.go
dm3ch: Also checked with LLM - cascade deletion is already seems handled correctly — the finalizer pattern is fully in place:
- `pkg/controllers/nodeclass/status/controller.go` adds `TerminationFinalizer` to every GCENodeClass on creation
- `pkg/controllers/nodeclass/termination/controller.go` blocks finalizer removal while any NodeClaim with `spec.nodeClassRef.name == nodeClass.Name` still exists (requeues every 10 min), matching AWS and Azure exactly

So deletion order is enforced: NodeClaims are gone before Kubernetes deletes the NodeClass object.

What `errNodeClassTerminating` handles is a narrower race that exists regardless of finalizer ordering: after DeletionTimestamp is set (finalizer not yet removed, NodeClaims still draining), in-flight reconciler calls for `IsDrifted`/`GetInstanceTypes` that were already queued will still arrive. Those need to return gracefully rather than panic.

dm3ch:
### E2E Test Results

**Cluster:** `europe-west4-a` | **Commit:** `1ff06134` | **Total duration:** `~20m` (parallel)

| Suite | Spec | Result | Duration |
|---|---|---|---|
| Consolidation | should consolidate an empty amd64 on-demand node | ✅ pass | 3m 37s |
| Drift | should replace a drifted amd64 on-demand node | ✅ pass | 5m 22s |
| Drift | should replace a node when GCENodeClass metadata changes | ✅ pass | 2m 27s |
| Expiration | should replace an expired amd64 on-demand node | ✅ pass | 4m 33s |
| GC | should delete an orphaned VM after its NodeClaim is force-removed | ✅ pass | 2m 59s |
| Networking | should provision a node with no external IP when enableExternalIPAccess is false | ✅ pass | 1m 26s |
| Provisioning | COS / amd64 / on-demand | ✅ pass | 1m 41s |
| Provisioning | COS / arm64 / on-demand | ✅ pass | 2m 30s |
| Provisioning | COS / amd64 / spot | ✅ pass | 2m 0s |
| Provisioning | COS / arm64 / spot | ✅ pass | 1m 59s |
| Provisioning | Ubuntu / amd64 / on-demand | ✅ pass | 2m 48s |
| Provisioning | Ubuntu / arm64 / on-demand | ✅ pass | 2m 36s |
| Provisioning | Ubuntu / amd64 / spot | ✅ pass | 2m 47s |
| Provisioning | Ubuntu / arm64 / spot | ✅ pass | 2m 37s |
| Repair | should replace a node whose KernelDeadlock condition has been True beyond the toleration | ✅ pass | 2m 52s |

**Overall: ✅ all pass** — 15/15 specs, no panics in Karpenter logs.

> This PR was prepared with AI assistance (Claude Code). All changes have been reviewed and verified by the author.

thameezb (approved):

PR #387, opened 2026-05-27
feat(proposal): local-ssd

### LINKED ISSUE #385: proposal for local-ssd support
Greetings! We'd like to contribute to karpenter-gcp. One of our use cases is local-ssd. That covers:

- configurable machine types like `n2` and `n2d` where you pick the count (N x 375 GiB)
- bundled machine types with a fixed count: `c4d-*-lssd`, `c4a-*-lssd` (N x 375 GiB), and `z3-*-highlssd` / `z3-*-standardlssd` (N x 3000 GiB)

A few issues with the existing localssd code:

1. Specifying `disks[N].category: local-ssd` fails to create a node with the GCE error: `Invalid value for field 'resource.disks[N].type': 'PERSISTENT'.  Cannot create local SSD as persistent disk.`
2. The current code doesn't distinguish Ephemeral from RawBlock. They're separate GKE modes. Ephemeral needs `NODE_LOCAL_SSDS_EPHEMERAL=true` in `kube-env`; RawBlock doesn't.
3. Setting the SSD count on the `GCENodeClass` is awkward. If you want different counts in the same pool you have to make one NodeClass+NodePool per count, identical except for the `disks[]` entries.

I have a branch that fixes the PERSISTENT vs SCRATCH bug and supports both Ephemeral and RawBlock. It also lets one NodeClass cover many machine types and SSD counts.

The count is set per pod via a label instead of on the NodeClass:

- `karpenter.k8s.gcp/instance-local-ssd-count: "<N>"`

For configurable families (`n1`, `n2`, `n2d`, `c2`, `c2d`) the scheduler emits one InstanceType variant per allowed count. For bundled SKUs (`c4-*-lssd`, `c4d-*-lssd`, `c4a-*-lssd`, `z3-*-{standard,high}lssd`, `a3-*`) the count is pinned by the machine type. The NodeClass only declares `localSsdMode`:

```yaml
kind: GCENodeClass
metadata: { name: ssd-pool }
spec:
  imageSelectorTerms: [{ alias: "ContainerOptimizedOS@latest" }]
  disks: [{ category: pd-balanced, sizeGiB: 50, boot: true }]

  # new attribute:
  localSsdMode: RawBlock   # or Ephemeral
```

Disks are attached at instance create as SCRATCH NVMe. The node's `kube-env` metadata is patched so GKE's existing bootstrap scripts handle format/RAID/mount. These are the same keys a native GKE node pool uses. Ephemeral mode adds `NODE_LOCAL_SSDS_EPHEMERAL=true`. RawBlock leaves the SSDs as raw NVMe block devices for workloads that handle their own format and mount.

`disks[].category: local-ssd` is deprecated. It stays in the CRD and karpenter logs a warning. Nobody can be using it successfully today since GCE rejects it, so deprecating it shouldn't break anyone.

Examples:

**Use case 1: configurable family, mixed counts in one pool**
```yaml
# NodePool requirements
- { key: karpenter.k8s.gcp/instance-family, operator: In, values: [n2, n2d] }

# pod A: 4x375 GiB
nodeSelector:
  node.kubernetes.io/instance-type: n2d-standard-8
  karpenter.k8s.gcp/instance-local-ssd-count: "4"

# pod B: 2x375 GiB
nodeSelector:
  node.kubernetes.io/instance-type: n2d-standard-4
  karpenter.k8s.gcp/instance-local-ssd-count: "2"

# pod C: no SSD
nodeSelector:
  node.kubernetes.io/instance-type: n2-standard-2
```

**Use case 2: bundled SKUs across families in one pool**

Bundled SSDs on c4/c4d/z3 require hyperdisk-balanced boot:

```yaml
kind: GCENodeClass
metadata: { name: ephemeral-ssd-pool }
spec:
  disks: [{ category: hyperdisk-balanced, sizeGiB: 50, boot: true }]
  localSsdMode: Ephemeral  # or RawBlock

# NodePool
- { key: karpenter.k8s.gcp/instance-family, operator: In, values: [c4, c4d, z3] }

# Pods
nodeSelector: { node.kubernetes.io/instance-type: z3-highmem-14-standardlssd }  # 1x3 TiB
nodeSelector: { node.kubernetes.io/instance-type: c4d-standard-8-lssd }         # 1x375 GiB
nodeSelector: { node.kubernetes.io/instance-type: c4-standard-2 }               # no SSD
```

For bundled SKUs the count is fixed by the machine type and can't be set.

**Use case 3: scheduling by ephemeral-storage request**

In Ephemeral mode each InstanceType variant publishes its local-SSD bytes as ephemeral-storage capacity. A pod that asks for storage without pinning an SSD count gets the smallest variant that fits:

```yaml
resources:
  requests: { ephemeral-storage: 2Ti }
nodeSelector:
  karpenter.k8s.gcp/instance-family: n2d
```

Karpenter picks an n2d SKU and SSD count with at least 2 TiB total. No per-pod count math.

**z3 onHostMaintenancePolicy fix**

This branch also fixes `onHostMaintenancePolicy` for z3. GCE requires `TERMINATE` for non-metal z3 SKUs with more than 18 TiB of bundled local-SSD (`z3-highmem-88-highlssd`, `z3-highmem-176-standardlssd`) and for bare-metal z3 SKUs. Smaller non-metal z3 SKUs require `MIGRATE`. The GCE API rejects the wrong policy.

## Test coverage

- 7 new provisioning specs across COS/Ubuntu, amd64/arm64, RawBlock/Ephemeral, pod-set count vs NodePool count, bundled vs configurable, no-SSD-label-in-mixed-pool
- 1 mixed-pool spec covering c4/c4d/z3 in a single hyperdisk-balanced pool
- A negatives suite that verifies no NodeClaim is created on invalid configs

## Current status and questions

All existing e2e tests pass. New local-ssd tests pass. I'll do more testing in our dev clusters soon.

- Any feedback on the approach?
- Is this a PR you'd accept?
- The PR is going to be large and I don't see a clean way to split it. The test coverage above is the main mitigation. Anything you'd want to see before a PR?

joemiller:
#### What type of PR is this?
/kind documentation

followup from #385 

improves local-ssd support

Fixes #385 

<details open><summary><h3>Greptile Summary</h3></summary>

This PR adds a proposal for Local SSD support in GCE NodeClasses. The main changes are:

- Adds `GCENodeClass.spec.localSsdMode` for RawBlock and Ephemeral modes.
- Models Local SSD count as a scheduler-visible instance type property.
- Documents same-name count variants, create-time count resolution, and reconciliation by name plus count.
- Describes disk attach, bootstrap metadata, z3 maintenance policy, tests, and migration steps.
</details>

<details><summary><h3>Confidence Score: 5/5</h3></summary>

This looks safe to merge.

- No blocking issues found in the changed code.

</details>

<details><summary><h3>Important Files Changed</h3></summary>

| Filename | Overview |
|----------|----------|
| proposals/0003-local-ssd-support.md | Adds the Local SSD support proposal with API, scheduling, create, reconciliation, metadata, testing, and migration guidance. |

</details>

<sub>Reviews (16): Last reviewed commit: ["Merge branch &#39;main&#39; into joem/local-ssd-..."](https://github.com/cloudpilot-ai/karpenter-provider-gcp/commit/73a91a7325c0737b5569827faa37d4e1bad2a94e) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=34160554)</sub>


proposals/0003-local-ssd-support.md
```diff
@@ -0,0 +1,228 @@
...
+- **No-SSD families** (e.g. `e2`): one variant, count requirement pinned to `0`.
+
+In `Ephemeral` mode, each non-zero variant's `ResourceEphemeralStorage` includes the local-SSD bytes (375 GiB × count for 375-GiB-partition families, 3000 GiB × count for z3). This is what makes scheduling by `requests.ephemeral-storage` work without an explicit count: Karpenter picks the smallest variant whose ephemeral-storage capacity satisfies the request.
+
+#### Count resolution at provisioning
+
+`resolveLocalSSDCount` (`pkg/providers/instance/instance.go`) picks the count to send to the GKE bootstrapper. Precedence:
+
+1. `MachineType.BundledLocalSsds.PartitionCount`: authoritative for bundled SKUs. Wins unconditionally. If the cache doesn't have the MachineType for a name that looks bundled, the resolver returns a retryable error so the Create loop falls through rather than booting a node with the wrong physical SSD count.
```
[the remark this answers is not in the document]
joemiller: c3/c3d support local SSDs only as bundled-SSD SKUs (the -lssd machine-name suffix - https://docs.cloud.google.com/compute/docs/general-purpose-machines#c3d-standard_1; they're already handled via the bundled-SSD path described in §Design Details, not the configurable allowlist. n4 and n4d don't accept local SSDs at all per GCP's machine-series support table (https://docs.cloud.google.com/compute/docs/disks/local-ssd). The five-family configurable list (n1, n2, n2d, c2, c2d) was verified against gcloud compute machine-types in May 2026; newer GCE families have shipped only as bundled-SSD SKUs.

proposals/0003-local-ssd-support.md
```diff
@@ -0,0 +1,228 @@
...
+| Operator uses an unsupported requirement operator (`NotIn`, `Gt`, etc.) on the SSD-count label | Medium | Provisioning fails | Resolver surfaces a `CreateError` with a clear message naming the unsupported operator. |
+| GKE changes the kube-env / kube-label keys used by the bootstrap script | Low | Local-SSDs wouldn't be formatted/mounted on Ephemeral; raw devices still appear on RawBlock | Keys are the same ones GKE-native pools set; verified against live `kube-env` dumps. Same risk profile as the existing `PatchKubeEnvForInstanceType`. |
+| Existing deployments rely on `disks[].category=local-ssd` | Low | Warning logged; legacy fallback resolver path still attaches SCRATCH disks on configurable families | GCE currently rejects this shape outright, so no working deployment can depend on it. On bundled SKUs the controller now refuses to schedule (clear error) instead of producing a node with too many SSDs. |
+| z3 capacity is regionally scarce | High | e2e tests may flake on uncapped runs | z3-touching specs gated behind `E2E_Z3_TESTS=true`, mirroring the existing `E2E_GPU_TESTS` pattern. |
+
+---
+
+## Test Plan
```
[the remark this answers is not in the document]
joemiller: Verified via `gcloud compute machine-types describe` (us-central1-a): both z3-highmem-88-highlssd and z3-highmem-176-standardlssd report bundledLocalSsds.partitionCount = 12; z3 partition size is 3000 GiB, giving 36,000 GiB (~35.2 TiB) per SKU — well above the 18 TiB (18,432 GiB) TERMINATE threshold. The examples in the proposal are accurate.

```console
$ gcloud compute machine-types describe z3-highmem-88-highlssd \
  --zone=us-central1-a \
  --format='value(bundledLocalSsds.partitionCount)'
12

$ gcloud compute machine-types describe z3-highmem-176-standardlssd \
  --zone=us-central1-a \
  --format='value(bundledLocalSsds.partitionCount)'
12
```

proposals/0003-local-ssd-support.md
```diff
@@ -0,0 +1,228 @@
+# Proposal: Local-SSD Support for GCE NodeClasses
+
+- **Status**: Draft
+- **Authors**: @joemiller
+- **Created**: 2026-05-26
```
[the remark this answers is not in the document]
joemiller: This one is a bit confusing, but it did at least point out that a 0002 proposal was added 2 hours ago. I'll rename this to 0003

proposals/0003-local-ssd-support.md
```diff
@@ -0,0 +1,228 @@
...
+- `TERMINATE` for non-metal z3 SKUs whose bundled SSD total exceeds 18 TiB (currently `z3-highmem-88-highlssd`, `z3-highmem-176-standardlssd`).
+- `MIGRATE` for all other z3 SKUs.
+
+The thresholds are derived from `MachineType.BundledLocalSsds.PartitionCount` × the partition size, not hardcoded SKU names, so future z3 variants don't need a code change.
+
+#### Code changes
+
+| File                                                     | Change                                                                                       |
+|----------------------------------------------------------|----------------------------------------------------------------------------------------------|
+| `pkg/apis/v1alpha1/gcenodeclass.go`                      | New `LocalSsdMode` field + `LocalSSDMode` enum. Hash version bump to `v4`.                   |
```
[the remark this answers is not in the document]
joemiller: Updated proposal to be more explicit here. 

> "PatchLocalSSDMetadata is a no-op when count is zero (no kube-env or kube-labels entries are written), so a zero-SSD node in a NodeClass with localSsdMode: Ephemeral never gets a stray NODE_LOCAL_SSDS_EPHEMERAL: \"true\" and bootstraps normally."

proposals/0003-local-ssd-support.md
```diff
@@ -0,0 +1,228 @@
...
+|-------------|-------------------------------------------|---------------------------------------------------------|
+| `RawBlock`  | `NODE_LOCAL_SSDS_EXT: "<count>,nvme,block"` | `cloud.google.com/gke-local-nvme-ssd=true`              |
+| `Ephemeral` | `NODE_LOCAL_SSDS_EPHEMERAL: "true"`        | `cloud.google.com/gke-ephemeral-storage-local-ssd=true` |
+
+The function removes the opposite-mode entries on each call, so flipping `localSsdMode` on an existing NodeClass produces a clean single-mode state on the next provisioning cycle. It returns an error if `kube-env` or `kube-labels` is missing from the metadata items, consistent with the existing `PatchKubeEnvForInstanceType` and `AppendGPUTaint` patterns.
+
+#### z3 `OnHostMaintenance` fix
+
+`onHostMaintenancePolicy` (`pkg/providers/instance/instance.go`) returns:
+
```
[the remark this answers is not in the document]
joemiller: Updated to explicitly state the formula. When mode=Ephemeral, the ephemeral space setting is set to total size of the ssd drives attached. Root is no longer used for kubelet ephemeral-storage in this case

Actual implementation (pkg/providers/instancetype/types.go):
- Ephemeral mode with SSDs: ResourceEphemeralStorage = totalSSDGiB only (boot disk excluded).
- RawBlock or no-SSD: ResourceEphemeralStorage = bootDiskGiB.

> +`PatchLocalSSDMetadata` is a no-op when count is zero (no kube-env or kube-labels entries are written), so the zero-SSD variant in a NodeClass with `localSsdMode: Ephemeral` never gets a stray `NODE_LOCAL_SSDS_EPHEMERAL: "true"` and bootstraps normally.

proposals/0003-local-ssd-support.md
```diff
@@ -0,0 +1,236 @@
...
+| z3 capacity is regionally scarce                                                               | High       | e2e tests may flake on uncapped runs                                                                | z3-touching specs gated behind `E2E_Z3_TESTS=true`, mirroring the existing `E2E_GPU_TESTS` pattern.                                                                                                        |
+
+---
+
+## Test Plan
+
+### Unit Tests
+
+- `pkg/providers/instance/resolve_localssd_count_test.go` — resolver precedence: bundled value wins; multi-value `In` and non-`In` operators return `CreateError`; legacy `disks[]` fallback; cache-miss retryable error.
+- `pkg/providers/instance/localssd_test.go` — `hasBundledLocalSSDs` for API-signal + name-fallback paths, including `-nolssd` siblings; `evaluateLocalSSDConflict` for legacy-disk + bundled-SKU.
```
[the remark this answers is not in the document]
joemiller: added

proposals/0003-local-ssd-support.md
```diff
@@ -0,0 +1,236 @@
...
+
+---
+
+## Summary
+
+Karpenter cannot currently provision GCE nodes with local SSDs. The existing `disks[].category: local-ssd` shape on `GCENodeClass` produces an API-level GCE rejection (`Cannot create local SSD as persistent disk`). It also does not distinguish GKE's two local-SSD exposure modes (Ephemeral vs RawBlock), and forces a 1:1 split of NodeClass/NodePool per SSD count when an operator wants mixed counts in a single pool.
+
+This proposal replaces that shape with a per-pod count label (`karpenter.k8s.gcp/instance-local-ssd-count`) plus a NodeClass-level mode (`spec.localSsdMode: RawBlock | Ephemeral`). Local SSDs are attached as SCRATCH NVMe at instance create, and `kube-env` is patched with the same keys GKE-native node pools set, so the in-tree GKE bootstrap scripts handle format / RAID / mount unchanged.
+
+Configurable families (`n1`, `n2`, `n2d`, `c2`, `c2d`) emit one InstanceType variant per allowed SSD count. Bundled-SSD SKUs (`c4*-lssd`, `c4a-*-lssd`, `c4d-*-lssd`, `z3-*-{standard,high}lssd`, `a2-ultragpu-*`, `a3-*`, `a4*-*`, `h4d-*`) pin the count from the machine type.
```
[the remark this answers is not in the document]
joemiller: fixed

proposals/0003-local-ssd-support.md
```diff
@@ -0,0 +1,236 @@
...
+## Future Direction
+
+- **Per-pod `localSsdMode`**: currently the mode is a NodeClass-level setting. A pod-side override (similar to the count label) would let a single pool serve both Ephemeral and RawBlock workloads, which doubles the InstanceType variant count.
+- **Honor pod `requests.ephemeral-storage` on RawBlock**: today only Ephemeral mode reports SSD bytes as `ephemeral-storage`. RawBlock workloads have to use the explicit count label. Reporting raw block bytes via a different resource (e.g. an extended resource) would let RawBlock workloads schedule by capacity too.
+
+---
+
+## Open Questions
+
+1. **Should the legacy `disks[].category: local-ssd` shape stay deprecated-with-warning, or be hard-removed?** Hard-removing would simplify the resolver precedence chain (drop step 3). The argument for keeping it one release is operator notice via the warning, but since the shape is a 100% GCE failure today, no operator has it set on a working cluster.
```
thameezb: If the current error rate is 100% and we currently hard fail if this is set, then it's safe to remove with a clear note in the migration guide.

thameezb (dismissed):
LGTM, well documented, clear intention and approach defined.

proposals/0003-local-ssd-support.md
joemiller: updated to remove the old local-ssd code paths and from the CRD, with a migration note

proposals/0003-local-ssd-support.md
```diff
@@ -0,0 +1,232 @@
...
+- **Created**: 2026-05-26
+- **Related Issues**: [#385](https://github.com/cloudpilot-ai/karpenter-provider-gcp/issues/385)
+
+---
+
+## Summary
+
+Karpenter cannot currently provision GCE nodes with local SSDs. The existing `disks[].category: local-ssd` shape on `GCENodeClass` produces an API-level GCE rejection (`Cannot create local SSD as persistent disk`). It also does not distinguish GKE's two local-SSD exposure modes (Ephemeral vs RawBlock), and forces a 1:1 split of NodeClass/NodePool per SSD count when an operator wants mixed counts in a single pool.
+
+This proposal replaces that shape with a per-pod count label (`karpenter.k8s.gcp/instance-local-ssd-count`) plus a NodeClass-level mode (`spec.localSsdMode: RawBlock | Ephemeral`). Local SSDs are attached as SCRATCH NVMe at instance create, and `kube-env` is patched with the same keys GKE-native node pools set, so the in-tree GKE bootstrap scripts handle format / RAID / mount unchanged.
```
dm3ch: Also `Local SSDs are attached as SCRATCH NVMe at instance create` as far as I understood from [GKE docs](https://docs.cloud.google.com/kubernetes-engine/docs/concepts/local-ssd#emptydir-volume) this is recommended option of usage, but if it's a default mode in GKE.

If not do we have any strong reason to make our default mode different from GKE default mode

proposals/0003-local-ssd-support.md
```diff
@@ -0,0 +1,232 @@
...
+    karpenter.k8s.gcp/instance-local-ssd-count: "<N>"   (configurable families)
+    node.kubernetes.io/instance-type: <bundled-sku>      (bundled-SSD SKUs)
+    → SCRATCH NVMe attached, kube-env patched, GKE bootstrap mounts as configured.
+```
+
+### API Changes
+
+**`GCENodeClass.spec.localSsdMode`** — new optional field, enum `RawBlock | Ephemeral`, default `RawBlock`. Set once per NodeClass. Controls only the mode; counts come from the scheduler.
+
+**`karpenter.k8s.gcp/instance-local-ssd-count`** — new well-known label key. Three roles:
```
dm3ch: karpenter.k8s.gcp/instance-local-ssd-count - what is pratical reason of introduction of a new label?
As far as I understand, neither Karpenter-AWS nor GKE, GKE NAP has simillar label.

So what makes our karpenter provider different from other solutions, so we need this uniq label and `cloud.google.com/gke-ephemeral-storage-local-ssd=true` is not enough for us?

Or this is some kind of thing that is needed for your platform and as nodepools are heterogeneous in Karpenter you have to distinct different local-ssd presets on pod level? If so should we destinc them by SSD count or also add SSD size or something else.

BTW, want to research latter if k8s have some pod resources or something like this mechanism that would allow to declare resource requirement without laveraging custom labels saying that SSD exists on node in a needed count

proposals/0003-local-ssd-support.md
```diff
@@ -0,0 +1,232 @@
...
+Before:
+  GCENodeClass.spec.disks: [{ category: local-ssd, ... }, ...]
+    → GCE: "Cannot create local SSD as persistent disk"
+
+After:
+  GCENodeClass.spec.localSsdMode: RawBlock | Ephemeral
+  pod.spec.nodeSelector:
+    karpenter.k8s.gcp/instance-local-ssd-count: "<N>"   (configurable families)
+    node.kubernetes.io/instance-type: <bundled-sku>      (bundled-SSD SKUs)
+    → SCRATCH NVMe attached, kube-env patched, GKE bootstrap mounts as configured.
```
dm3ch: `SCRATCH NVMe attached` as far as I understood from [GKE docs](https://docs.cloud.google.com/kubernetes-engine/docs/concepts/local-ssd#emptydir-volume) this is the recommended option of usage, but is it a default mode in GKE?

If not do we have any strong reason to make our default mode different from GKE default mode

proposals/0003-local-ssd-support.md
```diff
@@ -0,0 +1,232 @@
...
+
+### Design Details
+
+#### InstanceType variant emission
+
+`pkg/providers/instancetype/types.go` learns to emit the SSD-count requirement per variant:
+
+- **Bundled-SSD SKUs** (detected via `MachineType.BundledLocalSsds.PartitionCount` from the GCE API, with a name-suffix/-prefix fallback for cache-miss): one variant, count requirement pinned to the bundled value. Names ending in `-nolssd` (e.g. `c4d-highmem-8-nolssd`, `a3-ultragpu-8g-nolssd`) override the family-level bundled classification — they are treated as non-bundled with count pinned to `0`, so the prefix-based GPU-family match (`a3-`, `a4-`, etc.) does not over-promote them. The override applies to the name-fallback path only; when the GCE API is the source of truth, `BundledLocalSsds.PartitionCount` is already correct.
+- **Configurable-SSD families**: one variant per count returned by `AllowedLocalSSDCounts(machineName, vCPUs)`, plus a zero-SSD variant for pods that don't request any SSDs. (e.g. for `n2d-standard-8` the allowed counts are 1, 2, 4, 8, 16, 24, so 7 variants including zero.)
```
dm3ch: Am I right reading this as for machine types that are able for configurable local SSDs we are going to by default provide multiple options with different local ssd count?

If so we need to be sure if no pods requires local ssds we would preffer a machine with no local SSDs by default.
Not sure how should that be implemented. So far I have only idea of adding attached SSD price to machine price to preffer cheaper option (but I am not 100% sure that this is how karpenter works)

proposals/0003-local-ssd-support.md
```diff
@@ -0,0 +1,232 @@
...
+
+#### Count resolution at provisioning
+
+`resolveLocalSSDCount` (`pkg/providers/instance/instance.go`) picks the count to send to the GKE bootstrapper. Precedence:
+
+1. `MachineType.BundledLocalSsds.PartitionCount`: authoritative for bundled SKUs. Wins unconditionally. If the cache doesn't have the MachineType for a name that looks bundled, the resolver returns a retryable error so the Create loop falls through rather than booting a node with the wrong physical SSD count.
+2. The NodeClaim's `karpenter.k8s.gcp/instance-local-ssd-count` requirement (originating from `pod.spec.nodeSelector`, `pod.spec.affinity.nodeAffinity`, or `NodePool.spec.template.spec.requirements`; all three are normalized by Karpenter into the same `scheduling.Requirement` before reaching the resolver). Operator must be `In` with a single value. `NotIn` / `Gt` / `Lt` / multi-value `In` are surfaced as `CreateError` (scope cut: supporting them requires per-machine-type allowed-count metadata we don't carry).
+3. Zero.
+
+There is no legacy `disks[].category=local-ssd` fallback: the value is removed from the CRD enum (see API Changes), so the resolver never observes it.
```
dm3ch: We should defenetly add MIGRATION.md item for this hard deprecation even if we assume that this function never worked.

proposals/0003-local-ssd-support.md
```diff
@@ -0,0 +1,232 @@
...
+
+z3-touching specs are gated by `E2E_Z3_TESTS=true`, matching the existing `E2E_GPU_TESTS` convention.
+
+---
+
+## Acceptance Criteria
+
+The feature is complete when:
+
+- [ ] `spec.localSsdMode` is on `GCENodeClass` with CRD validation and a `v4` hash version.
```
dm3ch: As far as I understand in this case we can keep hash version the same if default value would be empty.

If default value is not equal empty then have to bump hash version as far as I understand.

BTW in both cases we need to ensure this field would cause drift (don't remember if now we have blacklist or whitelist for drifting)

dm3ch (commented):
Overall this proposal LGTM for me, but left few comments with questions.

Also I would propose you to think if implementation meant to be implemented in single PR or it would be better to plan few smaller scoped PRs and threat them as milestones of imolementation

proposals/0003-local-ssd-support.md
dm3ch: For example I found AWS simmilar label `karpenter.k8s.aws/instance-local-nvme` but it displays NVMe GBs and not NVME count, should we go the simmilar way?

proposals/0003-local-ssd-support.md
```diff
@@ -0,0 +1,232 @@
...
+  GCENodeClass.spec.localSsdMode: RawBlock | Ephemeral
+  pod.spec.nodeSelector:
+    karpenter.k8s.gcp/instance-local-ssd-count: "<N>"   (configurable families)
+    node.kubernetes.io/instance-type: <bundled-sku>      (bundled-SSD SKUs)
+    → SCRATCH NVMe attached, kube-env patched, GKE bootstrap mounts as configured.
+```
+
+### API Changes
+
+**`GCENodeClass.spec.localSsdMode`** — new optional field, enum `RawBlock | Ephemeral`, default `RawBlock`. Set once per NodeClass. Controls only the mode; counts come from the scheduler.
```
dm3ch: Should we consider that instances with more than one localSSDs can have different modes on different disks?

proposals/0003-local-ssd-support.md
dm3ch: Also need to analyse how it would work with UnavailableOfferings cache. As currently the key is instance type and zone, so if I right understand if we wouldn't be able to provision machine type with given attached local-ssds count it would block provision of this machine type in a given zone completely

proposals/0003-local-ssd-support.md
```diff
@@ -0,0 +1,232 @@
...
+
+- `pkg/providers/instance/resolve_localssd_count_test.go` — resolver precedence: bundled value wins; multi-value `In` and non-`In` operators return `CreateError`; cache-miss retryable error.
+- `pkg/providers/instance/instance_test.go` — `renderDiskProperties` appends SCRATCH NVMe entries for a configurable count and **none** for a bundled SKU (regression guard for the gcloud-matching body).
+- `pkg/providers/instance/localssd_test.go` — `hasBundledLocalSSDs` for API-signal + name-fallback paths, including `-nolssd` siblings.
+- `pkg/providers/metadata/localssd_test.go` — `PatchLocalSSDMetadata`: RawBlock and Ephemeral upserts; mode flip removes opposite-mode keys; missing `kube-env` / `kube-labels` → error.
+- `pkg/providers/instancetype/types_test.go` — variant emission: bundled SKU produces one pinned variant; configurable family produces `{0} ∪ allowed` variants; ephemeral-storage capacity includes SSD bytes in Ephemeral mode.
+
+### E2E
+
+Suite at `test/suites/local-ssd/`, three files, 14 specs total (9 positive, 5 failure-mode). Runs in roughly 7 minutes with `--procs=4`, z3 skipped by default (gated on `E2E_Z3_TESTS=true`).
```
dm3ch: In e2e we should check both COS and Ubuntu local ssd support, because as far as I understand that families have different startup scripts and we want to be sure, that localSSD support is tested on both OS

proposals/0003-local-ssd-support.md
joemiller: "SCRATCH NVMe attach" is not a mode it's the only physical attach mechanism for GCE local SSDs. Every local SSD on every machine attaches as SCRATCH NVMe regardless of RawBlock/Ephemeral

You're right about rawblock/ephemeral default though. We should match GKE's default. This is probably the mode most users would choose. It automatically handles 1 or more disks, formats and mounts to /var/lib/kubelet, and the space is immediately available to pods using emptyDir.

In our case we only use rawblock and manage disks with local-static-provisioner or topolvm (there are other options out there), but this is not a good default for most users.

proposals/0003-local-ssd-support.md
joemiller: "SCRATCH NVMe attach" is not a mode it's the only physical attach mechanism for GCE local SSDs. Every local SSD on every machine attaches as SCRATCH NVMe regardless of RawBlock/Ephemeral

You're right about rawblock/ephemeral default though. We should match GKE's default. This is probably the mode most users would choose. It automatically handles 1 or more disks, formats and mounts to /var/lib/kubelet, and the space is immediately available to pods using emptyDir.

In our case we only use rawblock and manage disks with local-static-provisioner or topolvm (there are other options out there), but this is not a good default for most users.

proposals/0003-local-ssd-support.md
joemiller: The label exists because of a GCP-specific axis that AWS and GKE NAP don't have:

- On AWS, local NVMe is fixed per instance type, so `instance-local-nvme` is purely descriptive. There's no count to choose.
- GKE NAP derives the count from the pod's ephemeral-storage request and provisions directly, so there's no static instance-type menu to disambiguate.
- The configurable GCP families (`n1/n2/n2d/c2/c2d`, plus `c3/n4/n4d` per your other comment) make the SSD count a creation-time choice. Karpenter selects from a static menu of instance types, so each (machineType × count) has to be its own variant, and `instance-local-ssd-count` is the requirement that distinguishes them.

That said, now that the default is `Ephemeral`, the label isn't the primary scheduling path. We advertise SSD capacity as `ResourceEphemeralStorage`, so pods schedule on `requests.ephemeral-storage`, and `cloud.google.com/gke-ephemeral-storage-local-ssd=true` is sufficient for the common case. The count label is the escape hatch for `RawBlock` (no capacity signal) and explicit pinning.

On a native k8s mechanism: agreed it's worth researching. As far as I know there's no portable way to request "N local devices" today, which is the `RawBlock` gap the label fills.

joemiller: `karpenter.k8s.aws/instance-local-nvme` reports aggregate GB because on AWS the devices are fixed and capacity is the only meaningful axis. I'd keep count for our case, for a few reasons:

1. GCP's native unit is count (`--local-ssd-count`, `NODE_LOCAL_SSDS_EXT: "<count>,nvme,block"`).
2. RawBlock consumers reason in number of devices (striping, separate NVME namespaces), not aggregate bytes.
3. z3 uses 3000 GiB partitions vs 375 GiB elsewhere, so a GB label would conflate very different device layouts. Count stays honest about topology.

Capacity-based selection is already covered by `ResourceEphemeralStorage` in Ephemeral mode, so the GB-style ergonomics exist there without overloading the label. We could add a descriptive GB label later for AWS parity, but I'd treat that as a nice-to-have rather than the selector.

dm3ch:
I haven't checked you recent editits after my comments but here's one more thing that bothering me.

I don't know how instance types with customisations should be handled by Karpenter and if there's any other provider that already made that (not on NodeClass level):
- From one point of few we can try create internal ephemeral instance types derivating common machine type (for example n2-standard-4-2ssd). This would allow to not to fight with InstanceType selection mechanism as we would just have additional instance types in the list and have to only check if requirements are met and sort by price.
- From another point of view introducing own internal instance types can be missleading for users, so we might want to display instance type for n2-standard-4 with 2 SSD as just n2-standard-4 for user convinience. But in this case we must somehow support provisioning instances with same instance type but different configuration and price (not depending on NodeClass config, as you proposal says).

Have you researched that already. Or maybe you already have some internal implementation of support additionally attached local SSDs to test hypotesys?

proposals/0003-local-ssd-support.md
joemiller: as mentioned on slack, but copying here for posterity, i initially tried adding ssd price to the configurable node types's ssd variants (eg: n2(d) and related older machines) but what would happen is that karpenter would immediately try to move the pod to a cheaper node (0 ssd variant) unless the nodepool was set to consolidate=WhenEmpty.

This along with the UnavailableOfferings cache is indeed challenging to get right. It's really only an issue with the older configurable ssd machine types, the newer types with a fixed count based on the machine-type are easier since it matches all of the assumptions baked into the original karpenter-aws scheduling core. I'm still thinking about how to handle this properly for all machine types and both RawBlock and Ephemeral modes.

dm3ch: Ok, sounds reasoable. We can use sorting, preffer mechanism and latter add price accurate calculations if needed/possible

thameezb (dismissed):
LGTM

joemiller:
hi all. The proposal has been rewritten. The original proposal turned out to be unworkable and fought with the karpenter core scheduling logic. A new proposal has been written after some new design + testing on an implementation branch w/ e2e tests. I believe the new proposal matches the behavior of upstream karpenter-aws with regards to local-ssd (or 'instance storage' in aws terms on ec2), and fits neatly into karpenter-core's existing scheduling logic. Related discussion on the k8s slack thread: https://kubernetes.slack.com/archives/C0B20K4KWP8/p1780085866627939

dm3ch (dismissed):
Overall LGTM. I think you can start implementation based on this proposal.

In the meantime, I wasn't able to doublecheck all technical details, so I assume we might face some small corrections of proposal during implementation.

@joemiller  As we discussed in Slack thread, I suggest adding rawblack resource to this proposal, as potential future direction and explanation why not now. It defenetly wouldn't be implemented now, but I believe it would be useful to document that, to be able to return this in future

joemiller:
thanks. TCovered in the [Future Work](https://github.com/cloudpilot-ai/karpenter-provider-gcp/blob/joem/local-ssd-proposal/proposals/0003-local-ssd-support.md#future-work) section. it documents a k8s-native RawBlock extended resource as a deferred future direction, why it isn't in scope now (the count label is a provisioning signal and the GCE create API still needs an explicit attach count), and the candidate mechanisms (Karpenter lifecycle hook vs. device plugin). I pushed a small wording tweak to name the RawBlock resource explicitly in the future work section. Let me know if you'd like it expanded further

dm3ch (approved):

PR #481, opened 2026-06-23
fix: avoid duplicate GKE kube-proxy on Karpenter nodes

dm3ch:
#### What type of PR is this?

/kind bug

#### What this PR does / why we need it:

Fixes duplicate GKE kube-proxy handling for Karpenter nodes:

- stops adding `node.kubernetes.io/kube-proxy-ds-ready`, so GKE's kube-proxy DaemonSet does not run alongside the node-owned static/mirror kube-proxy pod
- accounts for the static kube-proxy pod's 100m CPU request in provider overhead modeling
- strengthens provisioning e2e checks to require exactly one Running/Ready kube-proxy pod per node and no opposite-arch kube-proxy image

Live cluster validation showed the node-owned static kube-proxy manifest already uses the arch-independent `kube-proxy:<version>` image; the issue #487 symptom is addressed by preventing the extra DaemonSet kube-proxy path.

#### Related proposal (if applicable):

N/A

#### Which issue(s) this PR fixes:

Fixes #473
Fixes #487

#### Docs and examples

- [x] `docs/` and `examples/` updated (or this PR does not affect user-facing behaviour, configuration, or APIs)
- [x] `MIGRATION.md` updated (or this PR does not require any migration steps)

#### Special notes for your reviewer:

Validation:

```console
go test ./test/pkg/environment
go test ./pkg/providers/...
```

E2E on `karpenter-e2e-cluster` with PR image `ttl.sh/karpenter-provider-gcp-pr481-8f5fe571-24h:e2e-8f5fe571`:

```console
ginkgo --procs=4 --timeout=2h -v ./test/suites/{channel-image-selection,confidential,consolidation,drift,expiration,gc,kubelet_config,networking,provisioning,repair,storage}
```

All non-GPU suites passed across the run; `networking` was rerun separately with `--procs=1` after a transient OAuth token EOF in `BeforeSuite` and passed.

<details open><summary><h3>Greptile Summary</h3></summary>

This PR fixes duplicate kube-proxy pods on Karpenter-provisioned GKE nodes by removing `node.kubernetes.io/kube-proxy-ds-ready` from Karpenter's `WellKnownLabels` (preventing the DaemonSet from being scheduled via Karpenter's label injection) and accounting for the node-owned mirror pod's 100m CPU request as provider-level `SystemReserved` overhead so Karpenter's scheduling simulation reserves the right amount of capacity.

[...]

PR #491, opened 2026-06-25
fix: normalize kube-proxy image arch in kube-env

### LINKED ISSUE #487: kube-proxy daemonset architecture mismatch
Greetings!

First of all I wanna thank you all for your effort in supporting this amazing project!

I'm using this Karpenter provider for GCP since it's first release. Now that I'm migrating my nodes from x86_64 (amd64) to ARM64 (aarch64) I've found that kube-proxy DaemonSet is still being deployed in ARM nodes with the amd64 container image.

As GKE manages kube-proxy with addonmanager, it's impossible to edit it to use the multi-arch image.

I've tried to set the architecture taint in the nodepool and trying to avoid kube-proxy to be deployed in ARM nodes as it's no longer necessary in Dataplane V2, but in the `kube-env` GCE Instance Template the `KUBE_PROXY_DAEMONSET` var is set to `true` always.

Is there a way to tell Karpenter to set that variable to `false` for a specific NodePool? I don't think this is related to the GKE nodepool discovery, or is it related too?

Thanks.

dm3ch:
#### What type of PR is this?

/kind bug

Normalizes architecture-suffixed GKE `kube-proxy` image references in `kube-env` from `kube-proxy-{amd64,arm64}:<version>` to the multi-arch `kube-proxy:<version>` tag while applying the existing architecture bootstrap patch.

This keeps static/node-owned kube-proxy bootstrap metadata safe when the discovered GKE source nodepool architecture differs from the Karpenter target NodeClaim architecture. It complements the DaemonSet readiness-label work in #481 by covering the node-owned/static kube-proxy path instead of the addonmanager-managed DaemonSet path.

Implemented red-green TDD coverage for:

- arch-specific kube-proxy image normalization during amd64 -> arm64 patching
- normalization even when no `SERVER_BINARY_TAR_URL` is present
- already multi-arch kube-proxy images remaining unchanged

#### Related proposal (if applicable):

N/A

Fixes #487

#### Docs and examples

- [x] `MIGRATION.md` updated (or this PR does not require any migration steps)

Validation run locally:

```console
go test ./pkg/providers/...
```

### DIFF
--- a/pkg/providers/metadata/kubeenv_arch.go
+++ b/pkg/providers/metadata/kubeenv_arch.go
@@ -35,6 +35,9 @@ var (
 	serverBinaryArchRegex = regexp.MustCompile(`kubernetes-server-linux-(amd64|arm64)\.tar\.gz`)
 	// gkeReleaseVersionRegex extracts the GKE release version from a binary URL.
 	gkeReleaseVersionRegex = regexp.MustCompile(`/release/(v[^/]+)/`)
+	// kubeProxyArchImageRegex matches GKE kube-proxy image repositories with an
+	// architecture suffix. The suffix is stripped to use the multi-arch image tag.
+	kubeProxyArchImageRegex = regexp.MustCompile(`kube-proxy-(amd64|arm64)(:[A-Za-z0-9._+\-]+)`)
 )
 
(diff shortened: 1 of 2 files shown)


dm3ch:
Closing in favor of composing this fix into #481 together with the kube-proxy e2e coverage.

PR #563, opened 2026-08-13
feat(interruption): detect Spot VM preemption early via instance/preempted metadata

### LINKED ISSUE #295: feat(interruption): detect Spot VM preemption 2 minutes early via instance/preempted metadata
**What would you like to be added**:

Implement early Spot VM preemption detection using GCP's `instance/preempted` metadata endpoint with `preemptionNoticeDuration=120s`. This gives Karpenter a ~2-minute window to drain before shutdown, matching AWS SQS spot interruption warning behavior.

**Why is this needed**:

Karpenter currently learns about Spot VM preemption only after the instance receives the ACPI shutdown signal — by then, 2–10 s of the 30 s drain window is already consumed. GCP's `preemptionNoticeDuration=120s` (in Preview since April 2026) causes `instance/preempted` to flip 120 s before shutdown, giving a 2-minute warning identical to AWS's spot interruption notice. Without this, GCP Spot users have significantly less time to drain workloads gracefully compared to AWS.

**Proposed implementation**:

Deploy `node-problem-detector` as an optional Helm subchart (`npd.enabled: true` by default, can be disabled for clusters already running NPD). A custom plugin long-polls `instance/preempted` and sets a `GCESpotPreempting=True` node condition when preemption is imminent. The interruption controller reacts to this condition and begins drain immediately with `markUnavailable: true`.

- `charts/karpenter/Chart.yaml` — add `node-problem-detector` optional subchart
- `charts/karpenter/values.yaml` — add `npd.enabled` toggle and NPD values overrides
- `charts/karpenter/templates/npd-preemption-configmap.yaml` — ConfigMap with plugin script and monitor config
- `pkg/controllers/interruption/controller.go` — handle `GCESpotPreempting` condition
- `pkg/apis/v1alpha1/` — add `spec.preemptionNoticeDuration` field (enum: `0` or `120`) — wire into instance creation once the feature graduates from Preview to stable v1 API

The existing kubelet `KubeletNotReady / node is shutting down` path remains as a fallback.

**Note**: `preemptionNoticeDuration=120s` requires the `compute/beta` API and is currently in Preview. The NPD detection path provides value even at `0s` (default) — it still detects preemption 2–10 s earlier than the kubelet condition path.

**Completion requirements**:

This enhancement requires the following artifacts:

- [ ] Design doc
- [x] API change
- [x] Docs update

The artifacts should be linked in subsequent comments.

ryan-macdonald-nb:
#### What type of PR is this?

/kind feature
/kind api-change

Karpenter currently learns a Spot VM is going away only once kubelet reports the node
shutting down, which happens *after* the ACPI G2 Soft Off signal has been delivered. That
leaves a fraction of the 30s shutdown window for draining. This adds an earlier signal.

Three pieces:

1. **`GCENodeClass.spec.preemptionNoticeDuration`** (`0` or `120`) asks GCE to flip the
   `instance/preempted` metadata key ahead of the shutdown signal.
2. **An optional DaemonSet** running `node-problem-detector` with a single custom plugin
   that watches that key and sets a `GCESpotPreempting` node condition.
3. **Handling in the interruption controller** for that condition, checked ahead of the
   kubelet condition since the notice arrives first.

The existing kubelet path is unchanged and remains the fallback.

#### Related proposal (if applicable):

`proposals/0007-spot-preemption-notice.md` (included in this PR, status `Provisional`)

Fixes #295

---

#### Two corrections to the issue

**The `compute/beta` API is not needed.** The issue says `preemptionNoticeDuration` requires
`compute/beta` and defers the API field until it graduates. The field is already on the
stable client this repo vendors and uses — `Scheduling.PreemptionNoticeDuration` in
`vendor/google.golang.org/api/compute/v1/compute-gen.go`, and `pkg/providers/instance/instance.go`
imports `google.golang.org/api/compute/v1`. No beta package is vendored, so the API field
ships here rather than waiting.

**The detector defaults to off.** `docs/node-repair.md` documents that GKE Standard node
pools already run their own node-problem-detector, and node repair depends on the conditions
it sets. Defaulting to `true` would add a second one to every node of every existing install
on upgrade.

#### Three things I'd like a maintainer call on

These are called out as Open Questions in the proposal rather than settled unilaterally:

1. **Hand-written templates instead of an NPD subchart.** The issue asks for a subchart. The
   chart has no `dependencies:` block and no DaemonSet today, so a subchart means the first
   external chart-repo dependency, a `Chart.lock`, and a large third-party values surface
   that `values.schema.json` must declare in full (it sets `additionalProperties: false` at
   the root and in all nested objects). Three templates give the same behaviour with a much
   smaller surface. Happy to switch if you'd rather take the subchart — the plugin config
   and condition semantics carry over unchanged.
2. **`spotPreemptionNotice.enabled: false` by default** rather than the `true` in the issue,
   for the reason above.
3. **Whether the preemption path should mark the offering unavailable.** Implemented as
   `true`, following the issue and matching the AWS provider. Counter-argument: GCP also
   preempts Spot VMs simply for hitting their 24h maximum lifetime, which isn't a capacity
   signal, so treating every preemption as one may pull offerings out of rotation without
   cause.

#### Testing

- `make ci` passes (`verify-codegen`, `verify-deadcode`, `chart-lint`, `verify-crds`,
  `ut-test`, `docs-lint`); `golangci-lint run` reports 0 issues.
- Unit tests: `setupScheduling` across spot/on-demand and 0/120; controller deletes the
  NodeClaim, increments the disrupted counter, marks the offering unavailable, and publishes
  both events on `GCESpotPreempting=True`; no action when the condition is `False`.
- Chart: renders 0 DaemonSets by default and 1 when enabled. The rendered monitor config
  parses as JSON and the rendered plugin passes `sh -n` and `shellcheck`.

**Not verified:** the full 120s path against a real preemption. GCE decides when to preempt,
so that needs a live cluster running Spot capacity. Manual steps are in
`docs/spot-preemption.md`. I'd appreciate a second pair of eyes there if anyone has a Spot
workload handy.

#### Note for a follow-up (pre-existing, not touched here)

`cleanNodeClaimByInstanceName` returns an error when no NodeClaim matches the node, and that
error aborts the whole node loop for that reconcile. Once a NodeClaim is fully deleted but
the Node object lingers, `handleStoppingSpotInstances` errors every second until the Node
goes away. This behaviour is unchanged by this PR — the new path has the same shape as the
existing one — but it may be worth a separate fix.

## Summary by CodeRabbit

* **New Features**
  * Added configurable GCE Spot VM preemption notices with 0- or 120-second advance warning.
  * Added optional, disabled-by-default detection of Spot preemption signals.
  * Nodes receiving a notice are flagged, recorded as interrupted, removed from scheduling, and their capacity offering is temporarily marked unavailable.
  * On-Demand instances are unaffected, with shutdown detection retained as a fallback.

* **Documentation**
  * Added setup, configuration, troubleshooting, and reference guidance for Spot preemption notices.

#### Docs and examples

- [x] `docs/` and `examples/` updated
- [x] `MIGRATION.md` updated (not required — the new field defaults to `0` and the DaemonSet defaults to off, so upgrades are a no-op)

```release-note
Detect GCE Spot VM preemption up to two minutes early. Adds `GCENodeClass.spec.preemptionNoticeDuration` (`0` or `120`), an optional node-problem-detector DaemonSet that sets the `GCESpotPreempting` node condition (`spotPreemptionNotice.enabled`, off by default), and handling of that condition in the interruption controller. The existing kubelet shutdown path remains as a fallback.
```

<details open><summary><h3>Greptile Summary</h3></summary>

This PR adds early GCE Spot preemption detection and handling.
- Adds `GCENodeClass.spec.preemptionNoticeDuration` and passes it into Spot VM scheduling.
- Adds an optional node-problem-detector subchart that publishes `GCESpotPreempting`.
- Extends the interruption controller to delete affected NodeClaims and temporarily mark their offerings unavailable.
- Adds generated CRD, chart, test, proposal, and operator documentation updates.
</details>

<details><summary><h3>Confidence Score: 5/5</h3></summary>

The PR appears safe to merge because no blocking failure remains within the eligible follow-up-review scope.

No blocking failure remains.
</details>

<details><summary><h3>Important Files Changed</h3></summary>

| Filename | Overview |
|----------|----------|
| pkg/apis/v1alpha1/gcenodeclass.go | Adds the validated preemption-notice duration API field, which participates in NodeClass drift hashing when nonzero. |
| pkg/providers/instance/instance.go | Applies the configured notice duration to Spot VM scheduling while leaving on-demand scheduling unchanged. |
| pkg/controllers/interruption/controller.go | Detects the custom preemption condition before the kubelet fallback and initiates NodeClaim cleanup. |
| charts/karpenter/values.yaml | Configures the optional node-problem-detector monitor and metadata-server probe. |
| charts/karpenter/values.schema.json | Adds the dependency enablement value to the parent chart schema. |

</details>

<details open><summary><h3>Sequence Diagram</h3></summary>

```mermaid
sequenceDiagram
    participant GCE as GCE Metadata
    participant NPD as node-problem-detector
    participant K8s as Kubernetes API
    participant IC as Interruption Controller
    participant NC as NodeClaim
    GCE->>NPD: "instance/preempted = TRUE"
    NPD->>K8s: "GCESpotPreempting=True"
    IC->>K8s: Poll Node conditions
    IC->>NC: Delete matching NodeClaim
    IC->>IC: Mark Spot offering unavailable
```
</details>

<sub>Reviews (5): Last reviewed commit: ["fix(chart): keep the embedded probe scri..."](https://github.com/cloudpilot-ai/karpenter-provider-gcp/commit/e9a78d2593f0058f751a72f993e939d545d35878) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=53141946)</sub>

<details><summary><h4>Context used (4)</h4></summary>

- Knowledge Base — [Interruption Controller](https://app.greptile.com/karpenter-provider-gcp/-/custom-context/knowledge-base/cloudpilot-ai/karpenter-provider-gcp/-/docs/interruption-controller.md)
- Knowledge Base — [Instance Provider](https://app.greptile.com/karpenter-provider-gcp/-/custom-context/knowledge-base/cloudpilot-ai/karpenter-provider-gcp/-/docs/instance-provider.md)
- Knowledge Base — [GCENodeClass API Types](https://app.greptile.com/karpenter-provider-gcp/-/custom-context/knowledge-base/cloudpilot-ai/karpenter-provider-gcp/-/docs/gcenodeclass-api.md)
- Knowledge Base — [Helm Deployment and CRD Packaging](https://app.greptile.com/karpenter-provider-gcp/-/custom-context/knowledge-base/cloudpilot-ai/karpenter-provider-gcp/-/docs/helm-crd-packaging.md)
</details>


ryan-macdonald-nb:
Thanks — the `curl` finding was correct and it was a serious one. Pushed 932cfe5f.

### The bug

I confirmed it against the image rather than assuming:

```
$ docker run --rm --entrypoint /bin/sh \
    registry.k8s.io/node-problem-detector/node-problem-detector:v0.8.20 \
    -c 'for b in curl wget nc busybox python3 perl bash dash; do ...'
  absent curl
  absent wget
  absent nc
  absent busybox
  absent python3
  FOUND  perl
  FOUND  bash    (5.2.15, /dev/tcp supported)
  FOUND  dash
```

Worse than "the plugin errors out": my script did `result=$(curl ...) || result=""`, so a
missing `curl` produced an empty string, fell through every match, and **exited 0**. The
DaemonSet would have run happily, the condition would have sat at `False` forever, and no
preemption would ever have been detected. Silent.

### What changed

**1. Read the metadata server over bash's `/dev/tcp`.** No external binary needed. HTTP/1.0
with `Connection: close` means the server closes the socket after the body, so there's no
chunked encoding to parse.

**2. Split "not preempted" from "could not tell"** — this is the fix that actually matters.
The plugin now uses all three of node-problem-detector's exit codes:

| Exit | Meaning | Condition |
|------|---------|-----------|
| 0 | confirmed not preempted | `False` |
| 1 | confirmed preempted | `True` |
| 2 | unreachable / timeout / non-200 / unparseable | `Unknown` |

Any probe that can't get a trustworthy answer reports `Unknown`. That's what makes a
future breakage of this kind visible instead of silent — it's the root cause of why the
`curl` bug was invisible, not just the `curl` call itself.

**3. Dropped the `wait_for_change` long poll** for a 2s poll interval. Against a 120s
notice the extra lag is irrelevant, and it removes the pre-read that was only there to
avoid waiting out the timeout on an already-flagged node. Simpler and easier to reason about.

### Verification

Ran the **template-rendered** script (not a hand copy) inside the real
node-problem-detector image against a mock metadata endpoint:

| Case | Expected | Result |
|------|----------|--------|
| `TRUE` | 1 | pass |
| `FALSE` | 0 | pass |
| HTTP 500 | 2 | pass |
| unparseable value | 2 | pass |
| connection refused | 2 | pass |
| server hangs | 2 | pass |
| truncated response | 2 | pass |

Two bugs of my own surfaced while testing, both worth noting since they're easy to repeat
in shell: `$?` after a `while` loop is the loop's status rather than the last `read`'s, and
`while read` silently drops a final line that has no trailing newline — which was exactly
the response body. Both fixed; the truncated-response case exists to pin the first one.

Also confirmed the `#!/usr/bin/env bash` shebang resolves when the file is exec'd directly
at `0555`, which is how the ConfigMap mounts it.

The image constraint is now documented in `values.yaml` and `docs/spot-preemption.md`:
an `image.repository` override needs a bash with `/dev/tcp` support.

### Nits

Both applied — compound time expressions hyphenated, ASCII diagram fences tagged `text`.

charts/karpenter/values.yaml
```diff
@@ -163,6 +163,48 @@ controller:
...
+spotPreemptionNotice:
+  # -- Deploy the preemption notice detector. Off by default: GKE Standard node
+  # pools already run their own node-problem-detector, so enabling this adds a
+  # second one. It only lands on nodes matching `nodeSelector` below.
+  enabled: false
+  image:
+    # -- Repository for the node-problem-detector image running the detection plugin.
+    repository: registry.k8s.io/node-problem-detector/node-problem-detector
+    # -- Tag for the node-problem-detector image.
+    tag: v0.8.20
```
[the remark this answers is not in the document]
ryan-macdonald-nb: Confirmed and fixed in 932cfe5f — thank you, this was a real bug and a silent one.

Verified against the image directly: no `curl`, `wget`, `nc`, `busybox` or `python3`; only
`bash` (5.2.15, `/dev/tcp` supported), `dash` and `perl`.

The failure mode was worse than a plugin error. The script did
`result=$(curl ...) || result=""`, so the missing binary yielded an empty string, fell
through every match, and **exited 0** — the detector would have reported `False` forever
while every preemption went unnoticed.

Rather than change the image, the plugin now reads the metadata server over bash's
`/dev/tcp` (HTTP/1.0 + `Connection: close`, so no chunked encoding to parse), and — the
part that actually matters — distinguishes *not preempted* (exit 0) from *could not tell*
(exit 2 → `Unknown`). A probe that cannot get a trustworthy answer is now visible instead
of silent, which is the root cause of why this slipped through.

The constraint is documented in `values.yaml` and `docs/spot-preemption.md`: an
`image.repository` override needs a bash with `/dev/tcp` support.

Verification detail is in [this comment](https://github.com/cloudpilot-ai/karpenter-provider-gcp/pull/563#issuecomment-5286875110)
— 7 cases run against the template-rendered script inside the real NPD image.

docs/spot-preemption.md
```diff
@@ -0,0 +1,102 @@
+# Spot preemption notice
+
+GCE reclaims Spot VMs with little warning. By default Karpenter only finds out once the shutdown signal has already been delivered and kubelet reports the node as shutting down, which leaves a fraction of the 30 second shutdown window for draining.
```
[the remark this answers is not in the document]
ryan-macdonald-nb: Fixed in 932cfe5f — all four hyphenated (`30-second`, `120-second`, `24-hour`), plus one more instance of `120 second` further down the proposal that the finding did not list.

proposals/0007-spot-preemption-notice.md
```diff
@@ -0,0 +1,172 @@
...
+## Proposal
+
+### Overview
+
+Before — one signal, arriving after shutdown has begun:
+
+```
+GCE sends G2 Soft Off ──▶ kubelet marks node shutting down ──▶ Karpenter drains
+                          (already inside the 30s window)
+```
```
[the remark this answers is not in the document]
ryan-macdonald-nb: Fixed in 932cfe5f — both diagram fences tagged `text`.

charts/karpenter/templates/npd-preemption-configmap.yaml
```diff
@@ -0,0 +1,132 @@
...
+  {{- end }}
+data:
+  gce-preemption-monitor.json: |
+    {
+      "plugin": "custom",
+      "pluginConfig": {
+        "invoke_interval": "{{ .Values.spotPreemptionNotice.pollIntervalSeconds }}s",
+        "timeout": "{{ add .Values.spotPreemptionNotice.requestTimeoutSeconds 10 }}s",
+        "max_output_length": 120,
+        "concurrency": 1
```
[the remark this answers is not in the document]
ryan-macdonald-nb: Good catch — fixed in 761071bf. This is the same failure mode as the `curl` one, just at
startup instead of at runtime.

Verified against v0.8.20 rather than assuming. The field is real and defaults to `false`:

```go
// pkg/custompluginmonitor/types/config.go
defaultSkipInitialStatus = false
// SkipInitialStatus prevents the first status update with default conditions
SkipInitialStatus *bool `json:"skip_initial_status,omitempty"`
```

and `monitorLoop` calls `initializeStatus()` unless it is set, which publishes the default
conditions before any check has run.

Ran the real binary with the rendered config to confirm the fix takes effect:

```
custom_plugin_monitor.go:130] Skipping condition initialization:
  [{Type:GCESpotPreempting Status:False ... Reason:NoPreemptionNotice}]
```

That is the `skip_initial_status` branch — the condition is tracked internally but nothing
is published until a probe produces a result.

Worth noting for anyone reading later: with this set, `GCESpotPreempting` is simply absent
from the node until the first probe completes. That is fine for the interruption controller
— `node.GetCondition` returns a zero-value condition, which is not `True`, so it takes no
action, and the kubelet fallback path is untouched.

charts/karpenter/README.md
```diff
@@ -154,3 +154,15 @@ serviceMonitor:
...
+| spotPreemptionNotice.image.repository | string | `"registry.k8s.io/node-problem-detector/node-problem-detector"` | Repository for the node-problem-detector image running the detection plugin. The plugin reads the metadata server over bash's /dev/tcp, because this image ships no curl, wget or python. An override must provide bash with /dev/tcp support. |
+| spotPreemptionNotice.image.tag | string | `"v0.8.20"` | Tag for the node-problem-detector image. |
+| spotPreemptionNotice.nodeSelector | object | `{"karpenter.sh/capacity-type":"spot"}` | Which nodes run the detector. Spot capacity only by default, since on-demand instances are never preempted. |
+| spotPreemptionNotice.pollIntervalSeconds | int | `2` | How often to check the metadata server, in seconds. This bounds how late a preemption notice is noticed, so it should stay small relative to the 120s notice window. |
+| spotPreemptionNotice.priorityClassName | string | `"system-node-critical"` | PriorityClass for the detector. Should be high enough that the detector is not evicted from a node it is responsible for watching. |
```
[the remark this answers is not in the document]
ryan-macdonald-nb: Fixed in 761071bf. The README is generated by `helm-docs`, so the wording change went into the `values.yaml` comment and the README was regenerated via `make update`.

charts/karpenter-crd/templates/karpenter.k8s.gcp_gcenodeclasses.yaml
```diff
@@ -528,6 +528,21 @@ spec:
...
+                default: 0
+                description: |-
+                  PreemptionNoticeDuration is how many seconds before shutdown GCE flips the
+                  instance/preempted metadata key on a Spot VM. 0 (the default) gives no advance
+                  notice: the key flips at the same moment the ACPI G2 Soft Off signal is sent.
+                  120 gives a two-minute warning, letting Karpenter start draining before shutdown.
+                  Only applies to Spot capacity; ignored for on-demand nodes.
+                  Reading the notice requires an agent on the node that watches the metadata key
+                  and sets the GCESpotPreempting condition — see docs/spot-preemption.md.
+                enum:
```
thameezb, line 541: do you suppose we will enable support for any other values here? e.g somewhere between 0 and 120? else this is more complicated boolean switch, where a bool may be more appropriate (something like notifyPreemptionEarly etc)

charts/karpenter/values.yaml
```diff
@@ -163,6 +163,54 @@ controller:
     # Can be overridden per-NodeClass via GCENodeClass.spec.serviceAccount.
     defaultNodepoolServiceAccount: ""
 
+# -- Optional DaemonSet that watches the GCE metadata server for Spot preemption
+# notices and sets the GCESpotPreempting node condition, which Karpenter's
+# interruption controller reacts to. Pair it with
+# `GCENodeClass.spec.preemptionNoticeDuration: 120` for a two-minute drain window.
+# See docs/spot-preemption.md.
+spotPreemptionNotice:
```
thameezb: instead of adding our own resources here, would it be possible to rather install a sub chart? keeps it cleaner and allows for more customisation, also makes updates and upgrades cleaner

pkg/controllers/interruption/events/events.go
```diff
@@ -22,6 +22,21 @@ import (
 	"sigs.k8s.io/karpenter/pkg/events"
 )
 
+// PreemptionNoticeReceived records that GCE signaled an imminent Spot preemption
```
thameezb: Could comments only be added when things are not clear by the function name or definition, would prefer not to pollute the code base with comments

pkg/controllers/interruption/controller.go
```diff
@@ -44,6 +44,12 @@ const (
 	NodeConditionReasonKubeletNotReady = "KubeletNotReady"
 	NodeConditionMessageShuttingDown   = "node is shutting down"
 
+	// NodeConditionTypeGCESpotPreempting is set by a node-side agent that watches the
```
thameezb: Could comments only be added when things are not clear by the function name or definition, would prefer not to pollute the code base with comments

pkg/controllers/interruption/controller.go
```diff
@@ -84,8 +90,17 @@ func (c *Controller) handleStoppingSpotInstances(ctx context.Context) error {
 			continue
 		}
 
-		condition := node.GetCondition(&currentNode, corev1.NodeReady)
-		if condition.Status != corev1.ConditionTrue && condition.Reason == NodeConditionReasonKubeletNotReady && condition.Message == NodeConditionMessageShuttingDown {
+		// A preemption notice arrives ahead of the shutdown signal, so check it first.
```
thameezb: Could comments only be added when things are not clear by the function name or definition, would prefer not to pollute the code base with comments

pkg/controllers/interruption/controller.go
```diff
@@ -95,6 +110,22 @@ func (c *Controller) handleStoppingSpotInstances(ctx context.Context) error {
 	return nil
 }
 
+// isPreempting reports whether a node-side agent has flagged an imminent Spot
```
thameezb: Could comments only be added when things are not clear by the function name or definition, would prefer not to pollute the code base with comments

pkg/controllers/interruption/controller.go
```diff
@@ -95,6 +110,22 @@ func (c *Controller) handleStoppingSpotInstances(ctx context.Context) error {
...
+// preemption by setting the GCESpotPreempting condition.
+func isPreempting(n *corev1.Node) bool {
+	return node.GetCondition(n, NodeConditionTypeGCESpotPreempting).Status == corev1.ConditionTrue
+}
+
+// isShuttingDown reports whether kubelet has observed the node shutting down. This
```
thameezb: Could comments only be added when things are not clear by the function name or definition, would prefer not to pollute the code base with comments

pkg/controllers/interruption/controller_test.go
```diff
@@ -194,3 +195,100 @@ func TestHandleStoppingSpotInstances_DeletesShuttingDownKarpenterNode(t *testing
 	require.Equal(t, "spot-claim", kubeClient.deleted[0].GetName())
 	require.Equal(t, before+1, disruptedCounterValue(t, labels))
 }
+
+// preemptingNodeAndClaim builds a Karpenter-owned spot node carrying the given
```
thameezb: Could comments only be added when things are not clear by the function name or definition, would prefer not to pollute the code base with comments

pkg/providers/instance/instance.go
```diff
@@ -1034,14 +1034,22 @@ func (p *DefaultProvider) setupServiceAccounts(nodeClass *v1alpha1.GCENodeClass)
...
+// of the ACPI G2 Soft Off signal.
+func setupScheduling(capacityType string, nodeClass *v1alpha1.GCENodeClass) *compute.Scheduling {
 	sched := &compute.Scheduling{}
-	if capacityType == karpv1.CapacityTypeSpot {
-		sched.InstanceTerminationAction = instanceTerminationActionDelete
+	if capacityType != karpv1.CapacityTypeSpot {
+		return sched
+	}
+	sched.InstanceTerminationAction = instanceTerminationActionDelete
+	// Zero means no advance notice, which is also the GCE default, so leave the
```
thameezb: Could comments only be added when things are not clear by the function name or definition, would prefer not to pollute the code base with comments

pkg/providers/instance/instance_test.go
```diff
@@ -999,17 +999,53 @@ func TestSetupScheduling(t *testing.T) {
...
+	cases := []struct {
+		name            string
+		capacityType    string
+		noticeDuration  int64
+		expectedSeconds int64
+	}{
+		{name: "spot unset leaves notice duration nil", capacityType: karpv1.CapacityTypeSpot},
+		{name: "spot zero leaves notice duration nil", capacityType: karpv1.CapacityTypeSpot, noticeDuration: 0},
+		{name: "spot 120 sets a two-minute notice", capacityType: karpv1.CapacityTypeSpot, noticeDuration: 120, expectedSeconds: 120},
+		// GCE rejects preemptionNoticeDuration on non-preemptible instances, so it
```
thameezb: Could comments only be added when things are not clear by the function name or definition, would prefer not to pollute the code base with comments

proposals/0007-spot-preemption-notice.md
```diff
@@ -0,0 +1,179 @@
...
+## Future Direction
+
+- If GCE accepts notice durations beyond 120 seconds, the enum widens with no other change.
+- The unavailable-offerings TTL could be tuned per interruption cause rather than using the default.
+
+---
+
+## Open Questions
+
+1. **Should the preemption path mark the offering unavailable?** Implemented as `true`, following the issue and matching the AWS provider. The counter-argument: GCP also preempts Spot VMs simply for reaching their 24-hour maximum lifetime, which is not a capacity signal, so treating every preemption as one may pull offerings out of rotation without cause. Open — maintainer call.
```
thameezb, line 177: Maybe let's default this to false, and make it configurable to the user? based on their needs it could make sense to remove it from rotation, where as others may prefer that we don't

charts/karpenter-crd/templates/karpenter.k8s.gcp_gcenodeclasses.yaml
ryan-macdonald-nb: Fair question, and I do not have a strong view. Here is what I found so I am not guessing.

GCE documents exactly one opt-in value, `120s`, and the feature is still Preview. The API
field itself is a `Duration`, not an enum, so the wire format does not constrain it — but
nothing in the docs describes a range, a minimum, or a maximum. So you are right that today
this is a boolean with extra steps.

The argument for keeping the enum is only about later. If Google adds `60s` or `300s` while
this is in Preview, `Enum=0;120` widens to `Enum=0;60;120;300` and nothing breaks. A bool
would need deprecating and replacing with a duration field.

The argument for the bool is that it describes what actually exists today, and speculating
about values Google has not announced is how you end up with an awkward API either way.

Happy to go either way — it is your API and this is `v1alpha1`, so the cost of changing it
later is low. Tell me which you prefer and I will make the change.

charts/karpenter/values.yaml
ryan-macdonald-nb: Will do. I went with hand-written templates because the chart has no subchart dependencies
today and I did not want to be the first to add one, but I checked properly and your way
works fine: `settings.custom_monitor_definitions` mounts at `/custom-config` with
`defaultMode: 493`, so the probe script is executable, and `hostNetwork`, `nodeSelector`,
tolerations and `priorityClassName` are all values. My objection was weaker than I made it
sound.

Two things worth flagging while I do it.

The obvious chart is `deliveryhero/node-problem-detector`, which is third-party rather than
an official Kubernetes chart. That is a new kind of dependency for this repo. Say the word if
you would rather point at something else.

More importantly: **the version I pinned is stale.** I used `v0.8.20`; node-problem-detector
is on `v1.36.0` now and that chart defaults to `v1.35.1`. I will move to v1.x as part of this.

That last point matters for the `curl` fix in this PR — I verified "the image has no curl" and
"`skip_initial_status` defaults to false" against `v0.8.20` only. I will re-run both checks,
plus the whole probe matrix, against the v1.x image before I push, rather than assume they
carry across a major version.

proposals/0007-spot-preemption-notice.md
ryan-macdonald-nb: Happy to make it configurable. Two things before I pick an implementation.

**Where should the knob live?** You did not say, and the two options are meaningfully
different:

- A Helm value plumbed through to the controller, e.g.
  `controller.markPreemptedOfferingsUnavailable`, wired like the existing `FEATURE_GATES` env
  var. One operator-wide setting. This matches the fact that it is controller scheduling
  behaviour, and the unavailable-offerings cache is controller-global anyway.
- A field on `GCENodeClass`, so different node pools can differ. More flexible, but it is a
  second new API field in a PR you are already questioning the API shape of.

I lean to the Helm value for that reason, but it is your call.

**Worth surfacing:** this reverses the issue. #295 explicitly specifies `markUnavailable:
true`, which is why I implemented it that way. I would rather you and @dm3ch agree on the
default than have me implement one and get reversed by the other.

For what it is worth, I think defaulting to false is defensible on the merits: GCP preempts
Spot VMs for hitting their 24-hour maximum lifetime as well as under capacity pressure, so
treating every preemption as a capacity signal pulls offerings out of rotation for a reason
that may not exist. I raised that as an open question in the proposal.

pkg/controllers/interruption/events/events.go
ryan-macdonald-nb: Agreed, and this covers the other six comment threads too — I will strip them all rather than
reply on each.

You are right that most of them restate what the code already says. `isPreempting` and
`isShuttingDown` do not need a comment explaining that they check whether a node is
preempting or shutting down.

Two I would like to keep, and happy to drop either if you disagree:

- The `# --` comments in `values.yaml`. Those are helm-docs input and generate the chart
  README, so they are not really comments.
- The note that the node-problem-detector image ships no `curl`, which is why the probe talks
  to the metadata server over bash's `/dev/tcp`. That one is not obvious from the code, and
  the failure mode if someone "simplifies" it back to `curl` is silent — the probe exits 0
  and reports every node healthy forever. I will move it out of the script and into
  `docs/spot-preemption.md` instead, so the code stays clean and the reasoning is still
  somewhere findable.

charts/karpenter/values.yaml
ryan-macdonald-nb: Done in da629ac2. It is now the upstream chart as a subchart, gated on
`node-problem-detector.enabled` and still off by default. The monitor config and probe
script go in `settings.custom_monitor_definitions`, which the chart mounts at
`/custom-config` at mode 0755. My three templates are gone.

Two upstream defaults I overrode, both worth a look:

- The chart runs the pod **privileged** by default. Our plugin only needs network access, so
  it now runs unprivileged with a read-only root filesystem. I confirmed
  node-problem-detector still starts and runs the plugin that way.
- `settings.log_monitors` is emptied, so we only run the preemption plugin. The GKE-managed
  node-problem-detector already covers the kernel and container-runtime log monitors.

**On the version.** This moves off `v0.8.20` to the subchart's `v1.35.1`. I re-verified
rather than assuming my earlier findings carried across three majors — still no `curl`,
`wget`, `nc` or `python`, still bash 5.2 with `/dev/tcp`, `skip_initial_status` still
honoured. Full probe matrix re-run, and I confirmed end to end that node-problem-detector
flips `GCESpotPreempting` to `True` off the probe:

```
custom_plugin_monitor.go:284] New status generated:
  Conditions:[{Type:GCESpotPreempting Status:True Reason:SpotPreemptionImminent
               Message:instance is being preempted}]
```

**One thing that needs your call.** I vendored the chart tarball into
`charts/karpenter/charts/`. The release workflow runs `helm package` but never
`helm dependency update`, and without the tarball both `helm package` and `helm template`
fail outright. Vendoring keeps the workflow untouched and matches how the repo already
vendors Go dependencies, but it does commit an 8.5 KB binary. The alternative is adding
`helm dependency build` to `release-build.yaml` and `make chart-lint`. Happy to switch if you
prefer that.

Also worth knowing: the chart is only on `charts.deliveryhero.io` up to 2.3.14 / appVersion
v0.8.19. 2.4.1 is published as an OCI artifact at
`oci://ghcr.io/deliveryhero/helm-charts`, which is what the dependency points at.

pkg/controllers/interruption/events/events.go
ryan-macdonald-nb: All stripped in da629ac2, across all seven places. The no-curl explanation moved into `docs/spot-preemption.md` rather than being deleted, so the reasoning survives without sitting in the code.

docs/spot-preemption.md
```diff
@@ -0,0 +1,111 @@
...
+Without the detector, nothing reads the metadata key and setting `preemptionNoticeDuration` on its own changes nothing that Karpenter can see.
+
+## Enabling it
+
+Turn on the detector:
+
+```sh
+helm upgrade karpenter karpenter-provider-gcp/karpenter --install \
+  --namespace karpenter-system \
+  --set "node-problem-detector.enabled=true" \
```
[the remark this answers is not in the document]
ryan-macdonald-nb: I checked this one and I do not think it applies — the docs tree and the chart README use
different conventions, and this page follows the docs tree.

Every page under `docs/` uses `karpenter-provider-gcp/karpenter` and `karpenter-system`:

```
docs/getting-started/installation.md:92:  helm upgrade karpenter karpenter-provider-gcp/karpenter --install \
docs/node-repair.md:12:            helm upgrade karpenter karpenter-provider-gcp/karpenter --install \
docs/spot-preemption.md:23:        helm upgrade karpenter karpenter-provider-gcp/karpenter --install \
```

`namespace karpenter-system` appears 12 times across `docs/`, and no other namespace appears
at all. `docs/node-repair.md` is the closest sibling to this page — same shape, same feature-flag
framing — and this page matches it exactly.

The chart README does use `karpenter-gcp` and `--namespace karpenter`, so there is a real
divergence, but it is between the chart README and the whole docs tree and it predates this
PR. Changing just this page would make it the odd one out among the docs rather than fixing
anything. Happy to open a separate PR aligning the two if maintainers want that.

The other finding in this review was right and is fixed in e9a78d25 — the embedded probe
script contained `|` and `||`, which helm-docs rendered straight into the values table and
broke two rows.

charts/karpenter-crd/templates/karpenter.k8s.gcp_gcenodeclasses.yaml
dm3ch: GCP terraform according to the [doc](https://registry.terraform.io/providers/hashicorp/google/latest/docs/resources/compute_instance_template.html) handles `preemption_notice_duration` as a duration input with limited max values, as far as I udnerstood.

So I suggest using duration instead of enum just to keep interface familiar for GCP Terraform users.

P.S. In the meantime I understand that most of users would want to just use maximum available value here

charts/karpenter/values.yaml
dm3ch: Can you then please update workflows to work correctly with subchart?
Should we call dependency download before publish instead of storing dependency in repo?

dm3ch:
Finally found some time to review this PR.
Answering inital questions from PR body:
- Hand-written templates instead of an NPD subchart - makes sense for me as long as self-maintained manifests bring less maintenance burden than a subchart. 
- spotPreemptionNotice.enabled: false by default - sounds good for me
- Whether the preemption path should mark the offering unavailable - true is fine from my perspective. While note about 24 hours limit for spot instances looks missleading for me. If I right remember 24h limit had been part of GCP legacy spot offering and now they don't have expiration. (have found proofs in GPT, but wasn't able to find single official page claiming that, so maybe I am not right here)

Also created a separate issue from a note for followup - https://github.com/cloudpilot-ai/karpenter-provider-gcp/issues/616

proposals/0007-spot-preemption-notice.md
dm3ch: I checked other providers and haven't found any provider that allows switching preemption marking unavailable behavior, while different providers have chosen different behaviour about marking offering unavailable.

According to my research with ChatGPT it seems older GCP spot behaviour was to preempt the instance (even no capacilty pressure on GCP side) after 24 hours, but not for newer ones if I uderstood right - https://docs.cloud.google.com/compute/docs/instances/preemptible .
If so it means marking instance offering as unavailable after preemption is a correct behaviour, as preemption effectively means shortage of this instance type at given zone.

dm3ch (changes requested):
@ryan-macdonald-nb Thank you for contribution of this feature.

Overall LGTM, but I would request few improvements before merging.

Besides of what's left in comments discussion, can you please:
- Rebase your PR to actual main branch head
- Introduce e2e test spec, simmilar to NPDs. My idea of test, test sets `GCESpotPreempting` condition instead of our NPD and we check if we handle preemption condition right.
- Can you then please update workflows to work correctly with subchart? Should we call dependency download before publish instead of storing dependency in repo?

charts/karpenter-crd/templates/karpenter.k8s.gcp_gcenodeclasses.yaml
ryan-macdonald-nb: Done in a778437d. It is a duration now.

I read the Terraform docs you linked. `scheduling.preemption_notice_duration` is a duration object there: `seconds` (0 to 315,576,000,000) and `nanos` (0 to 999,999,999). That matches the GCE API, which models the field as a `Duration` proto. So you read it right.

I did not copy that nested shape. `nanos` means nothing for a notice GCE only documents at two minutes, and a nested object is clumsy to write in YAML. The field takes a duration string instead:

```yaml
spec:
  preemptionNoticeDuration: 120s   # or 2m
```

The Go type is `*metav1.Duration`. This API already uses it for `kubelet.evictionSoftGracePeriod`, and karpenter-core takes the same kind of duration string for `expireAfter` and `consolidateAfter`. It should read naturally from either direction.

Two things changed with it:

- **The default is unset rather than `0`.** Both leave the field off the GCE request. `0` only existed to make the enum work.
- **A CEL rule caps the value at 120s.** Anything above is rejected at admission.

**One judgement call I would like you to check.** The 120s cap is mine, not GCE's. GCE documents one value and the feature is still Preview, so I chose to fail at admission with a clear message rather than let GCE reject the instance later, where the error is slower and harder to read. The cost is that we reject a value GCE might start accepting. Raising it is one line:

```go
rule="duration(self) <= duration('120s')"
```

Say the word and I will drop the cap and let GCE decide.

### Verification

`make ci` passes. Unit tests cover `120s`, `2m` resolving to the same 120 seconds, `90s` passing through unrounded, `0s` and unset staying nil, and on-demand ignoring the field.

Against the generated CRD: `2m`, `0s` and `90s` are accepted; `121s` and `1h` are rejected with `preemptionNoticeDuration must not exceed 120s`; `500ms` and `abc` fail the pattern; a bare `120` is rejected as an integer.

One thing I checked because it would have bitten later: `metav1.Duration` re-serialises `120s` as `2m0s`. That still matches the pattern, so a controller rewriting the object cannot produce a value its own CRD rejects.

---

*Commit SHAs updated: the branch was rebased onto `main` to add DCO sign-offs, so the originals no longer exist.*

charts/karpenter/values.yaml
ryan-macdonald-nb: Done in fcfb9412. The archive is gone from the repo and the dependency is fetched at build time.

To answer your question directly: yes, download before publish is the better option. I vendored the tarball only because nothing in the repo resolved dependencies, so `helm package` failed without it. Wiring the resolution in is the smaller change, and it keeps an 8.5 KB binary out of git.

### What changed

- `charts/karpenter/charts/node-problem-detector-2.4.1.tgz` deleted, and the directory is gitignored.
- New `make chart-deps` target running `helm dependency build charts/karpenter`. `chart-lint` and `apply` now depend on it.
- A `Build chart dependencies` step in `release-build.yaml`, before `helm package`.

`Chart.lock` stays committed. I used `helm dependency build` rather than `helm dependency update` on purpose: `build` installs exactly the version and digest in the lock, while `update` re-resolves and rewrites it. So the release keeps pulling 2.4.1 at digest `sha256:7a1efb56...` until someone deliberately bumps it.

### Why chart-lint too, not just the release job

I expected the missing dependency to be a release-only problem. It is not, and the local failure is the worse of the two.

Without the archive, `helm package` fails loudly:

```
Error: found in Chart.yaml, but missing in charts/ directory: node-problem-detector
```

But `helm template --set node-problem-detector.enabled=true` exits 0 and renders **zero** DaemonSets. No warning. Anyone rendering the chart to check the detector would see it silently do nothing. `helm lint` only warns, so it would not have caught this either. Hanging the dependency off `chart-deps` means `make chart-lint` cannot hit that state.

### Verification

Cloned the branch fresh, so `charts/karpenter/charts/` does not exist, and ran the release steps:

| Step | Result |
|------|--------|
| `helm dependency build` | pulls 2.4.1, digest matches the lock |
| `helm package` | succeeds |
| packaged `.tgz` | contains the 17 subchart files |
| `helm template` on the packaged chart, detector enabled | 1 DaemonSet |

`make ci` passes. I also checked that `helm-docs` output is identical whether or not the archive is present, so `verify-codegen` does not become order-dependent on `chart-deps`.

One thing worth knowing: `helm dependency build` needs network access to `ghcr.io` during `make chart-lint`, which was not true before. The pull is anonymous and took about 4 seconds.

---

*Commit SHAs updated: the branch was rebased onto `main` to add DCO sign-offs, so the originals no longer exist.*

proposals/0007-spot-preemption-notice.md
ryan-macdonald-nb: You are right, and my earlier reply on this thread was wrong. Behaviour stays as it is: `markUnavailable: true`, not configurable. Proposal updated in 7509cafc; no code change, because it already did this.

I argued for making it switchable on the grounds that GCP preempts Spot VMs for reaching a 24-hour maximum lifetime, so a preemption would not always mean capacity pressure. I checked that claim against the GCE docs instead of trusting my memory of it, and it does not hold:

> preemptible VMs can only run for up to 24 hours at a time, but Spot VMs don't have a maximum runtime unless you limit the runtime

— [Preemptible VM instances](https://docs.cloud.google.com/compute/docs/instances/preemptible)

The 24-hour cap belongs to legacy Preemptible VMs. It is not a Spot VM behaviour, which is what this PR provisions. So the exception I built the argument on does not exist, and a Spot preemption does indicate pressure on that instance type in that zone. Marking the offering unavailable is the correct read.

Your point about no other provider making this switchable settles the rest of it. A knob here would be the only one of its kind, and it would need a default anyway.

I removed the claim everywhere it appeared rather than only from the Open Questions list, since it was also load-bearing in the risk table and the PR description. The proposal now records the decision and why the original argument was wrong, so it does not get re-litigated from the same bad premise later.

Thanks for pushing back on it — I would have shipped the wrong reasoning and left a knob in the API for it.

---

*Commit SHAs updated: the branch was rebased onto `main` to add DCO sign-offs, so the originals no longer exist.*

ryan-macdonald-nb:
Thanks for the review. All three points are addressed. Summary here; full detail in [this comment](https://github.com/cloudpilot-ai/karpenter-provider-gcp/pull/563#issuecomment-5768566929).

| Commit | Change |
|---|---|
| a778437d | `preemptionNoticeDuration` is a duration, not an `Enum=0;120` |
| fcfb9412 | Subchart archive fetched at build time instead of committed |
| 7509cafc | Proposal records the decisions and drops the bad 24-hour claim |

**Duration instead of enum.** The Terraform field you linked is a duration object with `seconds` and `nanos`, matching the GCE `Duration` proto. I used a duration string rather than that nested shape, so it is written `120s` or `2m`. The Go type is `*metav1.Duration`, which this API already uses for `kubelet.evictionSoftGracePeriod`. One judgement call there I would like you to look at: I capped it at 120s with a CEL rule, which is my decision rather than GCE's. [Original thread, now marked outdated](https://github.com/cloudpilot-ai/karpenter-provider-gcp/pull/563#discussion_r3781504455).

**Subchart dependency.** Downloading before publish is the better option, so that is what it does now. `helm dependency build` in a `chart-deps` make target and in the release job, with `Chart.lock` still committed so the version and digest stay pinned. I hung it off `chart-lint` as well, because without the archive `helm template` renders zero DaemonSets and exits 0 — the release failure is loud, the local one is silent. Verified from a fresh clone through to the packaged chart. [Original thread, now marked outdated](https://github.com/cloudpilot-ai/karpenter-provider-gcp/pull/563#discussion_r3781520856).

**The 24-hour claim.** You were right to doubt it and I was wrong to make it. GCE documents that "preemptible VMs can only run for up to 24 hours at a time, but Spot VMs don't have a maximum runtime unless you limit the runtime". The cap belongs to legacy Preemptible VMs, not Spot. That was the whole basis for treating a preemption as something other than a capacity signal, so marking the offering unavailable stays `true` and non-configurable. I have removed the claim from the proposal, the risk table and the PR description. [Original thread, now marked outdated](https://github.com/cloudpilot-ai/karpenter-provider-gcp/pull/563#discussion_r3781549839).

The proposal's Open Questions are now Resolved Questions, since all three are settled, and the sections that still described the first draft's hand-written templates now describe the subchart. Status moved from `Provisional`, which nothing else in `proposals/` uses, to `In Review`.

Thanks also for opening #616 for the `cleanNodeClaimByInstanceName` note.

CI is green on 7509cafc: Pre-submit, Docs and CodeQL.

**Still not verified:** the full 120-second path against a real preemption. GCE decides when to preempt, so it needs a live cluster running Spot capacity. Manual steps are in `docs/spot-preemption.md`.

---

*Commit SHAs updated: the branch was rebased onto `main` to add DCO sign-offs, so the originals no longer exist. That rebase also orphaned the inline review threads, so GitHub marks them outdated and hides them in Files changed — the detail is reposted in Conversation instead.*

ryan-macdonald-nb:
Branch is up to date with `main` again, and the SHAs changed — here is why.

`main` had moved 40 commits ahead. I brought the branch level and rebased it flat onto `main`, so every commit has a new SHA.

Two things I had to fix on the way:

- **DCO was red.** My three review-response commits had no `Signed-off-by`. That was my mistake; all commits are signed now.
- **`golangci-lint` rewrote one line.** `metav1.Duration` embeds `time.Duration`, so `notice.Duration.Seconds()` is the same as `notice.Seconds()`. The linter simplifies it, which left the tree dirty and failed `verify-codegen`. Folded into b437cf4e.

Nothing else changed. I checked the merged tree against the pre-rebase tree and they are identical, and `make update` produces no drift, so the generated CRDs and docs match their sources after picking up `main`'s controller-gen v0.22.0.

All checks green: presubmit, lint, Docs, CodeQL, DCO, Greptile.

ryan-macdonald-nb:
## Review responses, reposted here

I answered each of @dm3ch's points on its own inline thread. The rebase I did to add DCO sign-offs removed the commits those threads were anchored to, so GitHub now marks all three `Outdated` and hides them in **Files changed**. They are still in **Conversation**, but easy to miss.

Reposting the substance here, where it cannot go stale. Commit SHAs below are the current ones.

---

### 1. Duration instead of enum — a778437d

I read the Terraform docs you linked. `scheduling.preemption_notice_duration` is a duration object there: `seconds` (0 to 315,576,000,000) and `nanos` (0 to 999,999,999). That matches the GCE API, which models the field as a `Duration` proto. You read it right.

I did not copy that nested shape. `nanos` means nothing for a notice GCE only documents at two minutes, and a nested object is clumsy to write in YAML. The field takes a duration string instead:

```yaml
spec:
  preemptionNoticeDuration: 120s   # or 2m
```

The Go type is `*metav1.Duration`. This API already uses it for `kubelet.evictionSoftGracePeriod`, and karpenter-core takes the same kind of duration string for `expireAfter` and `consolidateAfter`.

Two things changed with it:

- **The default is unset rather than `0`.** Both leave the field off the GCE request. `0` only existed to make the enum work.
- **A CEL rule caps the value at 120s.**

**One judgement call I would like you to check.** The 120s cap is mine, not GCE's. GCE documents one value and the feature is still Preview, so I chose to fail at admission with a clear message rather than let GCE reject the instance later. The cost is that we reject a value GCE might start accepting. Raising it is one line:

```go
rule="duration(self) <= duration('120s')"
```

Say the word and I will drop the cap and let GCE decide.

Unit tests cover `120s`, `2m` resolving to the same 120 seconds, `90s` passing through unrounded, `0s` and unset staying nil, and on-demand ignoring the field. Against the generated CRD: `2m`, `0s` and `90s` are accepted; `121s` and `1h` are rejected with `preemptionNoticeDuration must not exceed 120s`; `500ms` and `abc` fail the pattern; a bare `120` is rejected as an integer.

One thing I checked because it would have bitten later: `metav1.Duration` re-serialises `120s` as `2m0s`. That still matches the pattern, so a controller rewriting the object cannot produce a value its own CRD rejects.

---

### 2. Subchart fetched at build time — fcfb9412

To answer your question directly: yes, download before publish is the better option. I vendored the tarball only because nothing in the repo resolved dependencies, so `helm package` failed without it.

- `charts/karpenter/charts/node-problem-detector-2.4.1.tgz` deleted, and the directory is gitignored.
- New `make chart-deps` target running `helm dependency build charts/karpenter`. `chart-lint` and `apply` depend on it.
- A `Build chart dependencies` step in `release-build.yaml`, before `helm package`.

`Chart.lock` stays committed. I used `helm dependency build` rather than `helm dependency update` on purpose: `build` installs exactly the version and digest in the lock, while `update` re-resolves and rewrites it.

**Why `chart-lint` too, not just the release job.** I expected this to be a release-only problem. It is not, and the local failure is the worse of the two. Without the archive, `helm package` fails loudly:

```
Error: found in Chart.yaml, but missing in charts/ directory: node-problem-detector
```

But `helm template --set node-problem-detector.enabled=true` exits 0 and renders **zero** DaemonSets. No warning. `helm lint` only warns, so it would not catch this either.

Verified from a fresh clone, where `charts/karpenter/charts/` does not exist:

| Step | Result |
|------|--------|
| `helm dependency build` | pulls 2.4.1, digest matches the lock |
| `helm package` | succeeds |
| packaged `.tgz` | contains the 17 subchart files |
| `helm template` on the packaged chart, detector enabled | 1 DaemonSet |

`helm-docs` output is identical whether or not the archive is present, so `verify-codegen` does not become order-dependent on `chart-deps`. One thing worth knowing: `helm dependency build` needs network access to `ghcr.io` during `make chart-lint`, which was not true before. The pull is anonymous and takes about 4 seconds.

---

### 3. Marking the offering unavailable — 7509cafc

You are right, and my earlier reply on that thread was wrong. Behaviour stays as it is: `markUnavailable: true`, not configurable. No code change, because it already did this.

I argued for making it switchable on the grounds that GCP preempts Spot VMs for reaching a 24-hour maximum lifetime, so a preemption would not always mean capacity pressure. I checked that against the GCE docs instead of trusting my memory:

> preemptible VMs can only run for up to 24 hours at a time, but Spot VMs don't have a maximum runtime unless you limit the runtime

— [Preemptible VM instances](https://docs.cloud.google.com/compute/docs/instances/preemptible)

The 24-hour cap belongs to legacy Preemptible VMs. It is not a Spot VM behaviour, which is what this PR provisions. So the exception I built the argument on does not exist, and a Spot preemption does indicate pressure on that instance type in that zone.

Your point about no other provider making this switchable settles the rest. I removed the claim everywhere it appeared, not just from the Open Questions list, since it was also load-bearing in the risk table and the PR description.

Thanks for pushing back on it — I would have shipped the wrong reasoning and left a knob in the API for it.

pkg/providers/instance/instance.go
```diff
@@ -1084,14 +1084,17 @@ func (p *DefaultProvider) setupServiceAccounts(nodeClass *v1alpha1.GCENodeClass)
...
-// configureInstanceCapacityProvision; this only wires the termination action so
-// GCE honors DELETE rather than the default STOP on preemption.
-func setupScheduling(capacityType string) *compute.Scheduling {
+// setupScheduling returns scheduling config derived from capacity type and the
+// NodeClass. Spot-specific fields (provisioning model, preemptibility) are set
+// later by configureInstanceCapacityProvision.
+func setupScheduling(capacityType string, nodeClass *v1alpha1.GCENodeClass) *compute.Scheduling {
 	sched := &compute.Scheduling{}
-	if capacityType == karpv1.CapacityTypeSpot {
-		sched.InstanceTerminationAction = instanceTerminationActionDelete
```
dm3ch, line 1094: why replacing `instanceTerminationActionDelete` with empty `sched`?

pkg/apis/v1alpha1/gcenodeclass.go
```diff
@@ -112,6 +112,19 @@ type GCENodeClassSpec struct {
...
+	// "120s" gives a two-minute warning, letting Karpenter start draining before shutdown.
+	// GCE currently accepts up to two minutes.
+	// Only applies to Spot capacity; ignored for on-demand nodes.
+	// Reading the notice requires an agent on the node that watches the metadata key
+	// and sets the GCESpotPreempting condition — see docs/spot-preemption.md.
+	// +kubebuilder:validation:Type=string
+	// +kubebuilder:validation:Pattern=`^([0-9]+(s|m|h))+$`
+	// +kubebuilder:validation:XValidation:message="preemptionNoticeDuration must not exceed 120s",rule="duration(self) <= duration('120s')"
+	// +optional
+	PreemptionNoticeDuration *metav1.Duration `json:"preemptionNoticeDuration,omitempty"`
```
ryan-macdonald-nb: @dm3ch — answering your duration point here, on the current code. The [original thread](https://github.com/cloudpilot-ai/karpenter-provider-gcp/pull/563#discussion_r3781504455) went outdated when I rebased for DCO, so it is hidden in Files changed.

**Done in a778437d.** It is a duration now.

I read the Terraform docs you linked. `scheduling.preemption_notice_duration` is a duration object there: `seconds` (0 to 315,576,000,000) and `nanos`. That matches the GCE API, which models the field as a `Duration` proto. You read it right.

I did not copy the nested shape. `nanos` means nothing for a notice GCE only documents at two minutes, and a nested object is clumsy in YAML. So it takes a duration string:

```yaml
spec:
  preemptionNoticeDuration: 120s   # or 2m
```

`metav1.Duration` is already used in this API for `kubelet.evictionSoftGracePeriod`, and karpenter-core takes the same kind of string for `expireAfter` and `consolidateAfter`. The default is now unset rather than `0`; both leave the field off the GCE request.

**One judgement call for you.** The 120s cap is mine, not GCE's. GCE documents one value and the feature is still Preview, so I chose to fail at admission with a clear message rather than let GCE reject the instance later. The cost is rejecting a value GCE might start accepting. Raising it is one line. Say the word and I will drop the cap.

Tests cover `120s`, `2m` resolving to the same 120 seconds, `90s` unrounded, `0s` and unset staying nil, on-demand ignoring it. Against the CRD: `121s` and `1h` are rejected with `must not exceed 120s`; `500ms` and `abc` fail the pattern; bare `120` is rejected as an integer.

Makefile
```diff
@@ -57,7 +57,10 @@ update: tidy download ## Update go files header, CRD and generated code
 verify-codegen: update ## Verify generated code is up to date
 	git diff --exit-code || (echo "Generated files are out of date — run 'make update' and commit the changes" && exit 1)
 
-chart-lint: ## Lint the Helm charts (validates values.schema.json and templates)
+chart-deps: ## Fetch Helm subchart archives pinned in charts/karpenter/Chart.lock
+	helm dependency build charts/karpenter
```
ryan-macdonald-nb: @dm3ch — answering your workflow question here, on the change itself. The [original thread](https://github.com/cloudpilot-ai/karpenter-provider-gcp/pull/563#discussion_r3781520856) went outdated when I rebased for DCO.

**Done in fcfb9412.** To answer directly: yes, download before publish is better. I vendored the tarball only because nothing in the repo resolved dependencies, so `helm package` failed without it.

- `charts/karpenter/charts/node-problem-detector-2.4.1.tgz` deleted, directory gitignored.
- This `chart-deps` target; `chart-lint` and `apply` depend on it.
- A `Build chart dependencies` step in `release-build.yaml`, before `helm package`.

`Chart.lock` stays committed. I used `helm dependency build` rather than `update` on purpose: `build` installs exactly the version and digest in the lock, `update` re-resolves and rewrites it.

**Why this hangs off `chart-lint` too, not just the release job.** I expected a release-only problem. The local failure is worse. `helm package` fails loudly:

```
Error: found in Chart.yaml, but missing in charts/ directory: node-problem-detector
```

But `helm template --set node-problem-detector.enabled=true` exits 0 and renders **zero** DaemonSets, with no warning. `helm lint` only warns, so it would not catch it either.

Verified from a fresh clone with no `charts/karpenter/charts/`: dependency pulls at the locked digest, `helm package` succeeds, the packaged `.tgz` carries the 17 subchart files, and templating that packaged chart with the detector enabled gives 1 DaemonSet. `helm-docs` output is identical either way, so `verify-codegen` does not become order-dependent on this target.

One thing worth knowing: this needs network access to `ghcr.io` during `make chart-lint`, which was not true before. The pull is anonymous and takes about 4 seconds.

proposals/0007-spot-preemption-notice.md
```diff
@@ -0,0 +1,189 @@
...
+## Future Direction
+
+- If GCE accepts notice durations beyond two minutes, the CEL bound widens with no other change.
+- The unavailable-offerings TTL could be tuned per interruption cause rather than using the default.
+
+---
+
+## Resolved Questions
+
+1. **Should the preemption path mark the offering unavailable?** Yes, and not configurable. Settled on review. An earlier draft argued the opposite on the grounds that GCP preempts Spot VMs for reaching a 24-hour maximum lifetime, which is not a capacity signal. That was wrong: the 24-hour cap belongs to legacy Preemptible VMs, and [GCE documents](https://docs.cloud.google.com/compute/docs/instances/preemptible) that "preemptible VMs can only run for up to 24 hours at a time, but Spot VMs don't have a maximum runtime unless you limit the runtime". Preemption of a Spot VM therefore does indicate pressure on that instance type in that zone, so marking the offering unavailable is correct. No other Karpenter provider makes this switchable.
```
ryan-macdonald-nb: @dm3ch — answering your offering-unavailable point here. The [original thread](https://github.com/cloudpilot-ai/karpenter-provider-gcp/pull/563#discussion_r3781549839) went outdated when I rebased for DCO.

**Recorded in 7509cafc.** You are right, and my earlier reply on that thread was wrong. Behaviour stays as it is: `markUnavailable: true`, not configurable. No code change, because it already did this.

I argued for a knob on the grounds that GCP preempts Spot VMs for reaching a 24-hour maximum lifetime, so a preemption would not always mean capacity pressure. I checked that against the GCE docs rather than trusting my memory:

> preemptible VMs can only run for up to 24 hours at a time, but Spot VMs don't have a maximum runtime unless you limit the runtime

— [Preemptible VM instances](https://docs.cloud.google.com/compute/docs/instances/preemptible)

The cap belongs to legacy Preemptible VMs. It is not Spot VM behaviour, which is what this PR provisions. So the exception I built the argument on does not exist, and a Spot preemption does indicate pressure on that instance type in that zone.

Your point about no other provider making this switchable settles the rest. I removed the claim everywhere it appeared, not just from this list — it was also load-bearing in the risk table and the PR description.

Thanks for pushing back on it. I would have shipped the wrong reasoning and left a knob in the API for it.

charts/karpenter-crd/templates/karpenter.k8s.gcp_gcenodeclasses.yaml
ryan-macdonald-nb: Note on this thread: GitHub marks it `Outdated` because I rebased onto `main` to add DCO sign-offs, which removed the commits it was anchored to. My answer above is unchanged and still current. It is also in [this Conversation comment](https://github.com/cloudpilot-ai/karpenter-provider-gcp/pull/563#issuecomment-5768566929) if the inline view hides it.

charts/karpenter/values.yaml
ryan-macdonald-nb: Note on this thread: GitHub marks it `Outdated` because I rebased onto `main` to add DCO sign-offs, which removed the commits it was anchored to. My answer above is unchanged and still current. It is also in [this Conversation comment](https://github.com/cloudpilot-ai/karpenter-provider-gcp/pull/563#issuecomment-5768566929) if the inline view hides it.

proposals/0007-spot-preemption-notice.md
ryan-macdonald-nb: Note on this thread: GitHub marks it `Outdated` because I rebased onto `main` to add DCO sign-offs, which removed the commits it was anchored to. My answer above is unchanged and still current. It is also in [this Conversation comment](https://github.com/cloudpilot-ai/karpenter-provider-gcp/pull/563#issuecomment-5768566929) if the inline view hides it.

pkg/providers/instance/instance.go
ryan-macdonald-nb: Good catch on the diff — it is misleading. Nothing was replaced, but I have changed it so the diff says that clearly. Done in 04d62d06.

`instanceTerminationActionDelete` was never dropped. I had inverted the condition into an early return:

```go
if capacityType != karpv1.CapacityTypeSpot {
    return sched
}
sched.InstanceTerminationAction = instanceTerminationActionDelete
```

Same behaviour on both paths — non-spot returns an empty `sched`, spot still sets DELETE. But because the guard moved, the termination action showed up as a `-`/`+` pair sitting next to a bare `return sched`, which reads exactly like what you asked about.

The restructuring bought nothing. The block is two statements. So I put it back to the original shape and nested the new line inside:

```go
sched := &compute.Scheduling{}
if capacityType == karpv1.CapacityTypeSpot {
    sched.InstanceTerminationAction = instanceTerminationActionDelete
    if notice := nodeClass.Spec.PreemptionNoticeDuration; notice != nil && notice.Duration > 0 {
        sched.PreemptionNoticeDuration = &compute.Duration{Seconds: int64(notice.Seconds())}
    }
}
return sched
```

The termination action is now untouched context and the diff shows only the added lines.

Both paths were already pinned by tests, which is why the behaviour change did not show up anywhere:

```go
t.Run("spot sets termination action", ...)        // requires instanceTerminationActionDelete
t.Run("on-demand leaves termination action empty", ...)
```

I also restored the part of the doc comment explaining why DELETE matters rather than the default STOP. I had cut it when the function stopped being only about the termination action, but the reason is still worth keeping.

dm3ch:
### E2E Test Results

**Cluster:** `asia-southeast1-b` | **Commit:** `a5e2dce8` | **Total duration:** `32m 9s`

|---|---|---|---|---|
| Channel Image Selection | ChannelImageSelection COS / channel: stable / amd64 / on-demand | ✅ pass | 2m 39s | |
| Channel Image Selection | ChannelImageSelection Ubuntu2404 / version: latest / amd64 / on-demand | ✅ pass | 4m 23s | |
| Confidential | Confidential should provision a SEV Confidential VM node | ✅ pass | 2m 32s | |
| Confidential | Confidential should provision a SEV_SNP Confidential VM node | ✅ pass | 3m 29s | |
| Consolidation | Consolidation should consolidate an empty amd64 on-demand node | ✅ pass | 3m 52s | |
| Consolidation | Consolidation should honor Empty disruption budgets for empty nodes | ✅ pass | 6m 41s | |
| Drift | Drift should replace a drifted amd64 on-demand node | ✅ pass | 3m 49s | |
| Drift | Drift should replace a node when GCENodeClass metadata changes | ✅ pass | 4m 13s | |
| Expiration | Expiration should replace an expired amd64 on-demand node | ✅ pass | 5m 12s | |
| Garbage Collection | GarbageCollection should delete an orphaned VM after its NodeClaim is force-removed | ✅ pass | 4m 50s | |
| Interruption | Interruption should replace a Spot node reporting GCESpotPreempting=True | ✅ pass | 3m 24s | |
| Kubelet Configuration | COS: systemReserved reduces allocatable | ✅ pass | 1m 52s | |
| Kubelet Configuration | COS: partial kubeReserved override keeps node Ready | ✅ pass | 1m 58s | |
| Kubelet Configuration | Ubuntu: systemReserved reduces allocatable | ✅ pass | 3m 54s | |
| Kubelet Configuration | COS: podsPerCore caps capacity | ✅ pass | 2m 2s | |
| Kubelet Configuration | COS: maxPods reflected in capacity | ✅ pass | 2m 3s | |
| Networking | Networking should provision a node with no external IP when enableExternalIPAccess is false | ✅ pass | 2m 27s | |
| Provisioning | Image Pinning should provision a node with a pinned ContainerOptimizedOS alias | ✅ pass | 2m 1s | |
| Provisioning | Image Pinning should provision a node with an exact image id | ✅ pass | 2m 32s | |
| Provisioning | Image Pinning should provision a node with a pinned Ubuntu alias | ✅ pass | 4m 9s | |
| Provisioning | Provisioning COS / amd64 / on-demand | ✅ pass | 2m 24s | |
| Provisioning | Provisioning COS / arm64 / on-demand | ✅ pass | 2m 18s | |
| Provisioning | Provisioning COS / amd64 / spot | ✅ pass | 2m 34s | |
| Provisioning | Provisioning COS / arm64 / spot | ✅ pass | 2m 39s | |
| Provisioning | Provisioning Ubuntu / amd64 / on-demand | ✅ pass | 3m 58s | |
| Provisioning | Provisioning Ubuntu / arm64 / on-demand | ✅ pass | 4m 0s | |
| Provisioning | Provisioning Ubuntu / amd64 / spot | ✅ pass | 4m 2s | |
| Provisioning | Provisioning Ubuntu / arm64 / spot | ✅ pass | 3m 57s | |
| Repair | NodeRepair should replace a node whose KernelDeadlock condition has been True beyond the toleration | ✅ pass | 3m 35s | |
| Storage | PDCSI Disk Type Labels should not schedule a pod that requires an unsupported disk type label | ✅ pass | 2m 19s | |
| Storage | PDCSI Disk Type Labels should schedule a pod that requires a supported disk type label before the node exists | ✅ pass | 2m 35s | |
| Storage | PDCSI Disk Type Labels should register disk type labels and topology for the selected machine family | ✅ pass | 2m 35s | |
| Storage | Disk Selection should override the N2 default with an explicit disk category | ✅ pass | 1m 46s | |
| Storage | Disk Selection should use the machine-series default when disk category is omitted | ✅ pass | 5m 35s | |

**Controller resources:** `requests cpu=500m memory=500MiB; peak cpu=145m memory=247.9MiB`

**Overall: ✅ all pass**

Controller log scan: no panics, fatals, permission, quota, or authentication errors. Cleanup-time missing NodePool/NodeClass reconciler errors and transient resource-version conflicts were observed; one transient cluster-sync wait was observed.

dm3ch (approved):

